PR #133, opened 2014-11-15
Introduce post_hook for filesystem storage

michaeladler:
This pull request adds the possibility to invoke external programs on new / updated local files, e.g. to post-process ics files and remove certain alerts.

Example:

```
[storage bob_calendar_local]
type = filesystem
path = ~/.calendars/
fileext = .ics
post_hook = /usr/local/bin/remove_alerts.sh
```

### DIFF
--- a/example.cfg
+++ b/example.cfg
@@ -46,6 +46,8 @@ fileext = .vcf
 # Create the directory if it doesn't exist: `True` or `False`
 #create = True
 #encoding = utf-8
+# For each new / updated file f, invoke the following script with argument f:
+#post_hook = /usr/local/bin/post_process.sh
 
 [storage bob_contacts_remote]
 type = carddav
@@ -69,6 +71,8 @@ collections = private,work
 type = filesystem
 path = ~/.calendars/
 fileext = .ics
+# For each new / updated file f, invoke the following script with argument f:
+#post_hook = /usr/local/bin/post_process.sh
 
 [storage bob_calendar_remote]
 type = caldav
--- a/tests/storage/test_filesystem.py
+++ b/tests/storage/test_filesystem.py
@@ -9,6 +9,7 @@
 '''
 
 import os
+import subprocess
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


untitaker:
@michaeladler BTW, did you find a way to achieve your goal without this functionality?

I want to implement concurrent item changes in vdirsyncer at some point (writing to multiple files at the same time, within the same collection), so I'm not sure if this makes sense then.

michaeladler:
@untitaker: No, I sticked to the current solution as it perfectly fits my use case. Building something that manually scans the folder and looks for changes is kinda like rebuilding the logic in vdirsyncer. One downside I see with my solution is the initial sync process as it launches an external program for every new ics file that comes in. After that, it's usually just a couple of ics files per day so I don't mind. Currently I have a cron job running vdirsyncer every 30 minutes.

untitaker:
What about diffing the file listing after synchronization?

untitaker:
@michaeladler I've been thinking about this quite a lot, and I think that a general purpose solution would be independent of the storage, i.e. work for singlefile storage too. Are you interested in writing this?

michaeladler:
@untitaker: Maybe. What kind of solution do you have in mind?

untitaker:
I thought about adding this parameter to the Storage baseclass, making it available in all storages. Inside the __init__ method we would then have to wrap the relevant methods.

michaeladler:
Let's talk about it from a high-level perspective. What should the behavior be? Launch the external program whenever we write something to disk (e.g. multiple .ics-Files or one single, large ics-File) and provide the file path as argument? I also thought about piping all the file names that have changed to the external program - this way only one process would have to be started (of course, this process might be xargs ;) )

untitaker:
Sorry for the late reply, was on mobile earlier.

Yeah, I didn't think too clearly about what we'd actually pass the commands. My initial intuition when writing that comment was that we'd simply invoke the command with the  ``href`` (internal identifier) of the updated item:

* for ``filesystem``, this would be the file's basename
* for CardDAV/CalDAV, this would be the path relative to the collection URL
* for ``singlefile`` and the other storages, this is basically something completely useless for the user. It's sometimes the UID of the item, sometimes a hash.

Whether we call a single command with all the files, or with only one at a time, I don't care.

Another idea would be not to call any commands, and just output machine-readable lines to stdout, which can be enabled via a special flag. It's up to the user to parse and handle these messages then.

----

I know I'm really indecisive on this topic, but I fear of extending the user interface of vdirsyncer so much it would be a pain to maintain. The idea that the exposed interface would at least be generally usable would somewhat comfort me, however.

untitaker:
@michaeladler In the end I decided to merge your PR. I suppose I have been overly eager with making everything "generally usable".

PR #161, opened 2015-01-05
Document certificate verification properly

untitaker:
@t-8ch, I added a testcase for ``verify_fingerprint`` which is not supposed to
fail IMO. Could you take a look at this?

- [x] ``verify=False, verify_fingerprint='foo'`` is an antipattern
- [x] https://github.com/shazow/urllib3/issues/529 -- just skipping test
- [x] comment at https://github.com/shazow/urllib3/pull/444 -- requiring newer version of requests
- [x] Debian actually modifies SSL


untitaker:
Also @eriolv, does Debian modify requests in any way to also use the system cert store?

untitaker:
Nevermind, requests version was old.

eriolv:
@untitaker sorry for the late reply, but yes, I use ca-certificates package, but using the vendor mechanism provided upstream:
http://anonscm.debian.org/viewvc/python-modules/packages/requests/trunk/debian/patches/01_use-system-ca-certificates.patch?revision=30954&view=markup

Cheers!

untitaker:
Thanks for the answer.

Issue #182, opened 2015-02-25
Support client-side certificates


mathstuf:
I've locked mine down (though I can allow non-cert auth again), but it'd be nice to have this.

PR #183, opened 2015-02-25
ssl: support client certificates

mathstuf:
---
Tested with Davical+nginx locally.

### DIFF
--- a/docs/ssl-tutorial.rst
+++ b/docs/ssl-tutorial.rst
@@ -53,4 +53,24 @@ the system certificate CAs. Normally these two stores are similar enough for
 you not to care. If the behavior on your system is somehow confusing, your best
 bet is explicitly setting the SSL options above.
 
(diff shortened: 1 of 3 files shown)


mathstuf:
Addresses #182.

untitaker:
Thanks for this patch!

* I think `auth_cert` is a more fitting name, because then all auth-related stuff starts with `auth_`. We also support JSON lists in the config, so only one param is needed, and the user can use both PEM and key/cert formats (see cert param in http://www.python-requests.org/en/latest/api/#requests.request)
* Please add an example to the docs you added, for both formats.
* Please also add this parameter to HttpStorage. There's also a shared docstring in there, ``HTTP_STORAGE_PARAMETERS``. Add some docs for that parameter there too.

mathstuf:
Updated.

untitaker:
Awesome, thanks!

PR #189, opened 2015-03-30
Mention todoman as a supported client.

hobarrera:
This is a bit of a shameless plug, but I *did* design todoman with vdirsyncer support in mind, and there doesn't seem to by anything else listed to cover these requirements. 😄

Thanks for vdirsyncer, BTW, it did make it possible for me to *finally* organize my list of todos.

### DIFF
--- a/docs/supported.rst
+++ b/docs/supported.rst
@@ -18,10 +18,15 @@ Calendars
   a calendar in a single ``.ics`` file. You can use
   :py:class:`vdirsyncer.storage.SingleFileStorage` with those.
 
+- todoman_, a CLI todo manager supports :doc:`vdir <vdir>`. It's designed to
+  manage todos (tasks), and not events. You can use
+  :py:class:`vdirsyncer.storage.FilesystemStorage` with it.
+
 .. _khal: http://lostpackets.de/khal/
 .. _dayplanner: http://www.day-planner.org/
 .. _Orage: http://www.kolumbus.fi/~w408237/orage/
 .. _rainlendar: http://www.rainlendar.net/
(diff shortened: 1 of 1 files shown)


untitaker:
Neat, thanks!

I'm currently using [watdo](https://github.com/untitaker/watdo) for this and like it, but others say that it's interface is kinky, so I'd always hoped that somebody would step up and write an alternative (I probably should mention watdo in the docs too).

I've found a minor issue with your current code, it's not writing the .ics-files [atomically](https://stackoverflow.com/questions/2705076/difference-between-writing-to-file-atomically-and-not). I've mentioned this in [the docs](https://vdirsyncer.readthedocs.org/en/stable/vdir.html), but don't really do a good job explaining it ATM. Take a look in [vdirsyncer's code](https://github.com/untitaker/vdirsyncer/blob/master/vdirsyncer/storage/filesystem.py#L144) to see how I implemented it.

hobarrera:
Never came across watdo before. It's definitely worth mentioning as well (people might be interested). Though I think event-oriented CalDav clients should be separated from todo-oriented clients if we start having too many.

Atomicity when writing slipped my mind, but I'll put it high up in my priority list.

Hope you enjoy the interface though. :)

hobarrera:
Huh, nice. I was already expecting to have to create a tmp-file, hardlink to it, and then clean up.

Looks like atomic_writer does all the work for me, thanks!

untitaker:
I've now restructured the docs.

I'm currently trying to get todoman running and have run into a few issues. Should I report those on the GitHub project or on your own one?

hobarrera:
Opening an issue on the [issue tracker](https://git.barrera.io/hobarrera/todoman/issues) is fine. I'll look into them ASAP.

PR #241, opened 2015-07-21
Factor out a few click-related things into extensions, new threading model

untitaker:
Changes:

* Almost all logging code is now done by `click-log`. The idea was to make that code available for `khal` and `todoman` as well. `khal` currently comes with what is basically a copy of vdirsyncer's logging code.
* A few generic multithreading-related things have been moved to `click-threading`. As a result, all user interaction is now done on the main thread, while worker threads put e.g. prompts on a queue instead of acquiring a lock. This fixes a bug where the terminal would be left in a broken state if the user hits `^C` on a password prompt.

- [x] Don't depend on dev packages.
- [x] Bug: password prompts get queued up if I hit `^C`.

At the same time I'm worried about introducing too many dependencies, because package maintainers tend to put each of those into a separate package. I wouldn't have anything against maintainers bundling all dependencies into a single package if that's more convenient. What do you think, @0-wiz-0, @hobarrera, @mbaldessari?


hobarrera:
`build.sh` (used for tests, IIRC), seems to install several dependencies from the latest `master`, but `setup.py` will install the latest stable via pip.

This might lead into issue in future where we depend on unreleased code. IMHO, travis should install via pypi too, so simulate what end-users (or packagers) will be doing.

untitaker:
Yeah the PR is not going to be merged like that. I'll have to wait until click 5 is released.

On 21 July 2015 21:07:41 CEST, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>`build.sh` (used for tests, IIRC), seems to install several
>dependencies from the latest `master`, but `setup.py` will install the
>latest stable via pip.
>
>This might lead into issue in future where we depend on unreleased
>code. IMHO, travis should install via pypi too, so simulate what
>end-users (or packagers) will be doing.
>
>---
>https://github.com/untitaker/vdirsyncer/pull/241#issuecomment-123446334

-- 
Sent from my phone. Please excuse my brevity.

hobarrera:
Okay, great.

Regarding the dependency introduction, I don't think it should be an issue. I very much prefer to depend on small libraries that do things well, rather than one huge blob of code. 😊

I don't think bundling them makes sense, especially since they'll be re-used by other applications that also interact with vdirs (khal, todoman).

0-wiz-0:
I'm fine with two more packages.
A related topic, showing my ignorance: the installed requires.txt notes what dependencies are needed, but I don't see warnings during "python setup.py build". It'd be great if that could warn about missing dependencies.
Am I doing something wrong or is that something that needs to be fixed in vdirsyncer (et al.)?

untitaker:
@0-wiz-0 I think you're not supposed to install the `.egg-info` directory. Take a look at https://aur.archlinux.org/packages/vd/vdirsyncer/PKGBUILD

Regarding warnings, I think it's better if you try to just run vdirsyncer from the built package to see if the package is missing any dependencies.

Please create a new issue or mail me in private if you have any questions regarding this though.

mbaldessari:
In general the policy for distros is to package each dependency separately (for a number of reasons I won't go into here). For me adding click-log and click-threading is not a problem per se. It will take me 
a bit of time to package them and get them reviewed, but nothing insormountable. Thanks for the heads-up ;)

untitaker:
Note that this might not even be merged for 0.6.0 though. Both `click-log` and `click-threading` are not very stable yet.

untitaker:
BTW do you guys honor the minimum version requirements strictly? Or should I add a runtime check to avoid confusion for users?

hobarrera:
On 2015-08-08 14:24, Markus Unterwaditzer wrote:
> BTW do you guys honor the minimum version requirements strictly? Or should I add a runtime check to avoid confusion for users?
> 

Are you talking about downstream packages?

Since Arch is bleeding edge, there's no dependency checking: the repos always
have the latest version, so vdirsyncer merely depends on the packages and
expects the latest version to be installed.

untitaker:
Perhaps the situation is a bit different with Arch, but I feel like there will
be some breakage if the user e.g. doesn't upgrade deps but upgrades vdirsyncer.

0-wiz-0:
I usually check dependencies, but please mention it in the NEWS if there are more important dependency pattern changes.
If done correctly, pkgsrc then handles it strictly :)

untitaker:
FWIW a simple run of installed vdirsyncer will uncover any dependency failures.
Perhaps this should be done in e.g. check() of a PKGBUILD.

Issue #245, opened 2015-07-27
Add option to verify fingerprint and certificate chain


mathstuf:
It would be nice to verify both the chain and the fingerprint. AFAICT, checking the fingerprint does not check the expiration date of the remote certificate (seeing as I just went through and regenerated certs which were being accepted even though they expired in April).

untitaker:
`verify_fingerprint` is currently also used in situations when the certificate is self-signed or somehow broken, yet otherwise the server supports SSL just fine.

I'm not sure how to allow this behavior without breaking this other usecase. Is this behavior really that useful?

mathstuf:
Why use the cert chain at all in that case? If both are there, both should have to pass though.

untitaker:
Because not everybody likes to or needs to pin by fingerprint. `verify_fingerprint` does disable the cert chain.

Also I imagine it's useful if you can use `verify = /path/to/custom` to pin against a particular set of certificates, or a particular CA (instead of all trusted CAs).

mathstuf:
Maybe there's some confusion here. Currently there are 4 ways to go:

  - neither `verify` nor `verify_fingerprint` -> error ("nothing to check against, why are you doing SSL?")
  - `verify` without `verify_fingerprint` -> certificate checking
  - no `verify` with `verify_fingerprint` -> fingerprint checking
  - `verify` and `verify_fingerprint` -> fingerprint checking

I'm just asking that the last do both certificate and fingerprint checks.

untitaker:
I understand. I assume that would also mean removing the behavior of automatically disabling `verify` if `verify_fingerprint` is set? Because otherwise we'd have to differentiate between "implicit verify" (no `verify` line in config) and "explicit verify" (`verify = True` in config), which is IMO confusing.

untitaker:
Fixed.

PR #252, opened 2015-08-06
ssl: use verify and verify_fingerprint

mathstuf:
Both have their uses. The latter is very strict in what it will accept,
but it does not catch expired certificates.

---
Addresses #245 without an option; set both, get both.

### DIFF
--- a/docs/ssl-tutorial.rst
+++ b/docs/ssl-tutorial.rst
@@ -16,9 +16,6 @@ To pin the certificate by SHA1- or MD5-fingerprint::
     ...
     verify_fingerprint = "94:FD:7A:CB:50:75:A4:69:82:0A:F8:23:DF:07:FC:69:3E:CD:90:CA"
 
-This disables :ref:`validation against root CAs <ssl-cas>`, since
-``verify_fingerprint`` is much stricter anyway.
-
 .. _ssl-cas:
 
 Custom root CAs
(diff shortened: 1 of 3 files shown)


untitaker:
Thanks!

t-8ch:
Did you test this? Urllib3 (The backend of requests) itself ignores the everything else if you give it a fingerprint.

untitaker:
>Did you test this? Urllib3 (The backend of requests) itself ignores the everything else if you give it a fingerprint.

The testsuite explicitly tests this. I think you changed this with https://github.com/shazow/urllib3/pull/444?

untitaker:
Ahh, I see now. Then the tests are broken...

t-8ch:
Yes. The quiestion was directed at @mathstuf.

untitaker:
Hmm, the current tests seem to run fine though. See https://github.com/untitaker/vdirsyncer/blob/97d65ab/tests/utils/test_main.py#L253-255

mathstuf:
I changed my local account to try and use the gmail cert I use for offlineimap and got errors. Adding back the fingerprint still got errors. It was late though; maybe I forgot a `:wq`? I'll test again tonight.

mathstuf:
Changing either the SSL cert or the fingerprint gives me errors, so a "production" test is indicating it is valid.

mathstuf:
Using urrllib3 1.10.2.

PR #302, opened 2015-12-12
improve doc on keyring usage

geier:


untitaker:
I get your point, but the user should really check which keyring backend was chosen for storing the password (which is why I removed feature of adding the password from vdirsyncer).

geier:
While I get your point (and you might want to add a warning to make people configure keyring and a link to keyring's documentation), I do believe the current situation is not really userfriendly.

Secondly, it will make people's systems less safe (because people are lazy, at least I am). If  people configure password retrieval wrongly and it doesn't work, theywill put their password into the config file in clear text (at least I did that after upgrading), which leads IMO to a much higher risk of accidentally publishing once password.

Also, if keyring returns None (what happens when it's not configured properly) a warning should be given to the user, instead of trying to use no password and than failing (I didn't try it for other password retrieval options).

If a username but no password is set, a warning should be printed as well (there is a warning if `auth` is set). If users really want to connect with a username but no password, they still can set an empty password (if that even is a thing).

untitaker:
>While I get your point (and you might want to add a warning to make people configure keyring and a link to keyring's documentation), I do believe the current situation is not really userfriendly.

Are you proposing adding the feature of "store password on prompt" back into vdirsyncer? I can't see a way how it'd work in a sensible way, given how the current "fetching" system is so generic.

Tbh this is precisely why I am getting sick of `keyring`, it adds an opaque abstraction layer that doesn't make it obvious where stuff is actually stored. One advantage of custom CLI commands is that you can just try them out on the console to see if they work. You don't have a true equivalent with `keyring`, unless you consider the Python REPL. But if you're not a Python dev, that's not a good experience.

Perhaps this could be circumvented with a new subcommand on `vdirsyncer` that allows users to get and set passwords using `keyring`, but it seems grossly out of scope for vdirsyncer.

>Also, if keyring returns None (what happens when it's not configured properly) a warning should be given to the user, instead of trying to use no password and than failing (I didn't try it for other password retrieval options).

I've fixed this by erroring out on all false-ish values for now. Since there's afaik no param in any vdirsyncer storage where an empty value makes sense, I think this doesn't kill any usecase, but oh well, we'll see.

geier:
Sorry for not replying to this before. No, I don't think the "save password feature" should be come back. Perhaps the keyring feature should just be removed, people who want to use it can use a minimal script like many people use it for offlineimap.

untitaker:
Opened #313 for further planning.

PR #303, opened 2015-12-13
Fix typo on collection discovery doc

JocelynDelalande:
(the example from the doc crashes on setting validation)


untitaker:
Thank you!

PR #306, opened 2015-12-16
Detect Debian-related problems and abort.

untitaker:


untitaker:
Since forever, we've been running into a lot of problems with Debian's packaging of requests. See https://twitter.com/untitaker/status/677152612353970176 for the high-level (subjective) overview, and https://github.com/kennethreitz/requests/pull/2567 (both discussion and the text in the diff) for the backstory. The first bug that was opened on vdirsyncer about this is https://github.com/untitaker/vdirsyncer/issues/82, but problems like it have continuously popped up ever since.

This is an attempt to detect such situations and give the user a more meaningful error message.

hobarrera:
I like this, since it's the most elegant I can think of solution to an upstream-inflicted problem. 👍

Issue #328, opened 2016-02-03
pair configuration: `collections = null` is not a good default.


untitaker:
`null` implies that the user entered direct URLs and paths to a collection. This is useful behavior in the case where collection discovery is completely broken on the server, but it is only rarely used and therefore a bad default. We have two options:

* Change the default, I suggest `["from a", "from b"]`
* Require a value, no default

lfam:
On Wed, Feb 03, 2016 at 02:05:29PM -0800, Markus Unterwaditzer wrote:
> `null` implies that the user entered direct URLs and paths to a
> collection. This is useful behavior in the case where collection
> discovery is completely broken on the server, but it is only rarely
> used and therefore a bad default. We have two options:
> 
> * Change the default, I suggest `["from a", "from b"]`

Since `vdirsyncer sync` syncs all configured pairs, then I think it
makes sense for this default to be to sync all collections from both
sides, which I think is what you are suggesting here.

untitaker:
>which I think is what you are suggesting here

Yes, but...

`vdirsyncer sync` syncs all *configured* pairs and collections, but the `collections` parameter defines the "configured collections". I'm not sure your analogy holds.

* All collections from both sides = All collections that are discoverable (through the DAV user account) or for which a local folder exists
* All configured collections = What is defined in `collections`.

lfam:
On Thu, Feb 04, 2016 at 11:33:08AM -0800, Markus Unterwaditzer wrote:
> >which I think is what you are suggesting here
> 
> Yes, but...
> 
> `vdirsyncer sync` syncs all *configured* pairs and collections, but the `collections` parameter defines the "configured collections". I'm not sure your analogy holds.
> 
> * All collections from both sides = All collections that are discoverable (through the DAV user account) or for which a local folder exists
> * All configured collections = What is defined in `collections`.

You're right, that is different.

untitaker:
In #331 I've decided to explicitly require the collections parameter for now. We may re-add a default later on, but I fear that simply changing the default would break user's configuration in too subtle ways.

untitaker:
0.9 is released without default, next release will be with new default.

untitaker:
Removing from 1.0 milestone since re-adding a default is not a breakage. However, I wonder if we ever should readd a default at all.

PR #331, opened 2016-02-11
Remove the default of collections to autodiscovery

untitaker:
See #328


untitaker:
cc @hobarrera @geier @lfam

geier:
The question really is: what is most people's use case. I believe this is ['from a', 'from b']. But I do agree with you, that changing to a completely different default is better done after the next release.

untitaker:
Yes, I'd make `["from a", "from b"]` the default for 1.0 (or later).

Issue #344, opened 2016-02-27
http calendars fail in 0.90


clauseggers:
```
vdirsyncer discover
PAIRS ()
Discovering collections for pair my_icloud_contacts
Discovering collections for pair my_icloud_calendars
Discovering collections for pair my_facebook_events
Discovering collections for pair type_events
Discovering collections for pair dutch_holidays

Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/vdirsyncer/cli/utils.py", line 217, in _discover_from_config
    discovered = list(cls.discover(**config))
  File "/usr/lib/python3.4/site-packages/vdirsyncer/storage/base.py", line 102, in discover
    raise NotImplementedError()
NotImplementedError

Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/vdirsyncer/cli/utils.py", line 217, in _discover_from_config
    discovered = list(cls.discover(**config))
  File "/usr/lib/python3.4/site-packages/vdirsyncer/storage/base.py", line 102, in discover
    raise NotImplementedError()
NotImplementedError
critical: Failed to initialize my_facebook_events_remote
critical:   - 

Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/vdirsyncer/cli/utils.py", line 217, in _discover_from_config
    discovered = list(cls.discover(**config))
  File "/usr/lib/python3.4/site-packages/vdirsyncer/storage/base.py", line 102, in discover
    raise NotImplementedError()
NotImplementedError
critical: Failed to initialize dutch_holidays_remote
critical:   - 
critical: Failed to initialize type_events_remote
critical:   - 
Saved for my_icloud_calendars: collections = ["home", "work", "tasks", "83027CAB-E781-4BAB-86EA-C59D68EFD890", "2EE90DA0-1B1A-4CAA-A478-A45969F27E6B", "EAFBAC39-A823-4E09-B0AA-FA0C900C97CA", "F4988EF2-55B1-4688-884F-F3CB3895D87A"]
Saved for my_icloud_contacts: collections = ["card"]
```

```
vdirsyncer sync
Discovering collections for pair type_events
Syncing my_icloud_calendars/home
Syncing my_icloud_calendars/2EE90DA0-1B1A-4CAA-A478-A45969F27E6B
Syncing my_icloud_calendars/work
Syncing my_icloud_calendars/F4988EF2-55B1-4688-884F-F3CB3895D87A
[...]

PR #346, opened 2016-02-27
Fix coerce_native

untitaker:
See #344


vdirsyncer/cli/utils.py
```diff
@@ -18,7 +18,7 @@
 from .. import DOCS_HOME, exceptions
 from ..sync import IdentConflict, StorageEmpty, SyncConflict
 from ..utils import expand_path, get_class_init_args
-from ..utils.compat import to_native
+from ..utils.compat import PY2
```
hobarrera: Unused import?

untitaker: Ah right.

PR #354, opened 2016-03-02
Add packaging guidelines

untitaker:
Fix #352

Issue #352, opened 2016-03-01
Packaging guidelines


untitaker:
I think it would make sense to document a few things for packaging vdirsyncer.

## How to test vdirsyncer

Many packagers try to run the full testsuite on a vdirsyncer installation, in order to find bugs with their setup. But: 99% of all packaging-related bugs are caught by seeing if `vdirsyncer --help` works correctly.

The testsuite is not suitable for packaging tests because:

* It uses [hypothesis](http://hypothesis.readthedocs.org/) to randomly generate example input data. Automatic test generation is a really powerful tool for finding encoding-related bugs, but it also means that tests may randomly fail.

* It launches CardDAV/CalDAV servers for integration testing. In other words, it's not just for finding bugs in vdirsyncer, but also elsewhere. Why would vdirsyncer-packagers care if a server introduced a bug that makes the tests fail?

* It requires very recent dependencies. Older pytest, hypothesis or Radicale versions may fail with bizzarre errors. This is especially a problem if packagers try to use their system-packages for installing test dependencies instead of just using a virtualenv and pip (I don't see a reason why that would be a problem TBH) See also #334 

This seems to be the sanest approach to testing vdirsyncer as a packager: https://github.com/Homebrew/homebrew/pull/41208 i.e. just create an example config and run a simple sync. Even a simple `vdirsyncer --help` catches 80% of errors IMO.

## Updating vdirsyncer

[The Hypothesis Packaging Guidelines](https://hypothesis.readthedocs.org/en/release/packaging.html) say that package maintainers who are unable to keep up with its release cycle are doing the community a disservice. While I think there are nicer ways to phrase that (especially towards people who are likewise doing a voluntary service), I am worried that vdirsyncer will find itself in a similarly problematic situation once [Debian starts to package it](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=745027). Packages usually have been very quickly updated otherwise.

Apart from that we should probably provide an RSS feed for new releases. GitHub has a tags RSS feed that is mentioned [in the docs](https://vdirsyncer.readthedocs.org/en/latest/changelog.html), I guess we could just use that.

## Enforcing dependencies' minimum required versions

Vdirsyncer has a lower bound on the versions of some dependencies. As far as I understand those lower bounds have been completely ignored so far by downstream. Perhaps there's just no way to fix this situation. Again I'm more worried about Debian than any other distro here, but maybe it'll be fine.

--

Seeking feedback from

[...]

PR #356, opened 2016-03-04
Allow running deterministic tests

hobarrera:
Make tests deterministic when the `DETERMINISTIC_TESTS` environment variable is set to true. This satisfies requirements of downstream packagers which require deterministic tests.

Also run larger test samples when running on CI, in hope that at some point it might trigger a new edge case.

See #352
Would probably require an update to #354

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,10 +3,11 @@
 General-purpose fixtures for vdirsyncer's testsuite.
 '''
 import logging
+import os
 
 import click_log
-
 import pytest
+from hypothesis import settings, Verbosity
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


untitaker:
Could you add a default variable to the `Makefile`, and don't fallback to a default in Python code (just to be consistent with other envvars)?

tests/conftest.py
```diff
@@ -27,3 +28,16 @@ def benchmark():
...
+    max_examples=1000,
+    verbosity=Verbosity.verbose,
+))
+settings.register_profile("deterministic", settings(
+    derandomize=True,
+))
+
+if os.getenv('DETERMINISTIC_TESTS', False):
+    settings.load_profile("deterministic")
+elif os.getenv('CI', False):
```
untitaker: Please compare both variables to `"true"`.

hobarrera:
Fixed both issues.

untitaker:
Merged into #354, thanks @hobarrera!

PR #357, opened 2016-03-04
Implement command to print minimal requirements

hobarrera:
This allows using `python setup.py requirements` to print minimum requirement versions.

Use `python setup.py --quiet requirements` to avoid setuptools printing "running requirements", which we don't want to see when scripting.

**NOTE**: this PR intends to merge into `packaging-guidelines`, and *not* into `master`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,9 +11,51 @@ Vdirsyncer is a synchronization tool for vdir. See the README for more details.
 
 
 import platform
+import re
 
-from setuptools import find_packages, setup
+from setuptools import find_packages, setup, Command
 
+
+requirements = [
+    # https://github.com/mitsuhiko/click/issues/200
+    'click>=5.0',
+    'click-log>=0.1.3',
+    'click-threading>=0.1.2',
+    # https://github.com/kennethreitz/requests/issues/2930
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,10 +11,55 @@
...
+    user_options = []
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        rog = re.compile("([^,]*)")
```
untitaker: I don't get why we need this. `requests==2.0.1, != 2.9.0` should parse fine?

setup.py
```diff
@@ -11,10 +11,55 @@
...
+        [
+            print(rog.match(
+                requirement.replace(">", "=").replace(" ", "")
+            ).group(1))
+            for requirement in requirements
```
untitaker: Please use a for-loop here :)

setup.py
hobarrera: You're absolutely right. I don't know why I though pip was dumber.

hobarrera: Fixed. Wow, that was ugly and unnecessary!

PR #370, opened 2016-03-09
Clean up logging usage

### LINKED ISSUE #369: Remove `vdirsyncer.log.get`
It is replaced by `click_log` and shouldn't be necessary anymore. Also the fact that other modules don't just use `logging.getLogger` was because I misunderstood how the logging module works.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31724879-remove-vdirsyncer-log-get?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hobarrera:
Fixes #369

### DIFF
--- a/vdirsyncer/cli/__init__.py
+++ b/vdirsyncer/cli/__init__.py
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
 
 import functools
+import logging
 import sys
 
 import click
 
 import click_log
 
-from .. import __version__, log
+from .. import __version__
 from ..utils.compat import PY2
 
 
-cli_logger = log.get(__name__)
+cli_logger = logging.getLogger(__name__)
 
 
 class AppContext(object):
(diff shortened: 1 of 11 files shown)


vdirsyncer/cli/__init__.py
```diff
@@ -38,7 +39,7 @@ def inner(*a, **kw):
 
 
 def validate_verbosity(ctx, param, value):
-    x = getattr(log.logging, value.upper(), None)
+    x = getattr(logging.getLogger, value.upper(), None)
```
untitaker: Not quite. :)

untitaker: concretely, it should be just `logging` since `value` is a log level.

hobarrera: It looks like this function is never called though, so is it safe to delete it? Or did I miss something?

untitaker: Oh, yeah...!

untitaker:
@hobarrera style build fails, please run `make install-style style` locally to see issues.

hobarrera:
`flake-import-order` seems odd, because it doesn't sort things like `isort` does (which is, alphabetically, and `from...` imports first).

Fixed though.

untitaker:
Always up for discussion. I sometimes get it wrong too.

untitaker:
Thanks!

PR #387, opened 2016-03-14
Minimal PR build, v2

### LINKED ISSUE #362: Travis CI is awfully slow
With the current amount of servers tested, CI takes **two and a half hours** to complete.

I'm not sure how to get around this, but the solution I'm thinking about involves *two* CI services, one that only tests `make DAV_SERVER=radicale test` and `make style`, and one that test the whole range. The second one would be called only for master or so.

I'm aware of the following CI services that can be integrated with GitHub:

* https://drone.io/
* https://circleci.com/
* Travis itself

For the full build, we absolutely need root access, and the ability to add arbitrary third-party PPAs. As far as I understand, this is a feature that Travis, Drone and Circle CI all have.

A setup using two services is pretty unfortunate, however. I'd prefer a solution that allows one to define a different build matrix for PRs than for pushes to master or tags.

--

This supersedes #350 and probably depends on #361 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31599514-travis-ci-is-awfully-slow?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github).
</bountysource-plugin>

untitaker:
Fixes #362


untitaker:
This is the final solution.

* I've refactored the Travis config generator to emit JSON instead of YAML, since that's easier to generate with Python's stdlib.
* Most Travis jobs are a no-op for pull requests: They will do absolutely nothing.
* It's still `homu`'s job to test the PR against the full matrix once I ping it to. However, I'll probably skip this for most PRs that don't touch any DAV stuff.

hobarrera:
LGTM.
And yes, the JSON generating code is a lot more readable.

PR #391, opened 2016-03-18
Split recurring events properly

untitaker:
Previously we moved each VEVENT into its own Item, now we group by UID,
if one exists.


untitaker:
@homu r+

untitaker:
@homu r-

untitaker:
@homu r+

Issue #8, opened 2014-03-06
Google OAuth for DAV


untitaker:
https://developers.google.com/google-apps/calendar/auth

Apparently we have to register the client at Google, not sure how we would keep the keys secret (if any) in an opensource project.

vaskas:
Is the default token really needed? Looking at https://github.com/insanum/gcalcli/blob/master/gcalcli#L642, you just need to call `OAuth2WebServerFlow` from `oauth2client.client` to do the browser dance.

untitaker:
Yes, one has to call it with client-id and -secret. I myself don't have any incentive to work on this though.

untitaker:
Somebody mailed me in private and hinted that adding OAuth support to this is now easier since remoteStorage support is now merged into master, which already uses OAuth.

* First of all, OAuth can't even be described as a proper standard. Many things can be described "OAuth-compatible", and I suspect that Google's OAuth implementation is vastly more complex than the one remoteStorage prescribes.

  Concretely, remoteStorage requires us to implement what is called the "implicit grant". flow. Basically you acquire one `access_token`, and that token will be used for any further authentication at all times.

  However, other variants of OAuth may require us to acquire a token to request another token that may be rotated every n days. For reasons I haven't looked into, this enhances security in case any credentials are stolen, but it's obviously harder to implement.

  It also means that if we implement OAuth for Google, this form of OAuth probably can't be reused for other services that require OAuth too, since no two OAuth implementations look alike.

* Google's CalDAV implementation is allegedly *okay*, while CardDAV seems to be abysmal. See http://evertpot.com/google-carddav-issues/ (already posted above).

* As already outlined above, we'd have to register vdirsyncer as an application with Google, and then provide *another token* to Google to identify ourselves as that application. We have a problem here: Should we let the user "play developer" and register their own application, and then let vdirsyncer "impersonate" that application? Or should we ship with a default token?

    * If the user registers their own application, using vdirsyncer with Google becomes somewhat hard to use.

    * If we ship with default token, anybody can easily steal that token and use it in their own apps.

  I think we should ship with a default token registered by me or whoever would maintain Google OAuth support.

Personally I won't be implementing this. I really don't care about supporting vendor-specific auth flows this complex.

PR #394, opened 2016-03-26
OAuth2 (Google variant) for CalDav and CardDav

marmarek:
This is mostly implementation of the ideas in https://github.com/pimutils/vdirsyncer/issues/8

It adds new `auth` value of `oauth2_google`. Then reuse `password` setting for token storage path. Actual authentication code must be obtained by the user, by following printed URL and then pasting retrieved code into the console. That happens only for the first time, later stored token is used.

Open questions, todos:
 - request and fill google app token (client id, client secret)
 - add option to specify them in configuration - I don't know how to get a (global?) config value from inside of Storage object
 - write some tests? but how to do that without exposing test calendar credentials (you can't simply run local google calendar instance...)? then how to handle interactive google confirmation? 
 - interactive prompt for authentication code (only at first run) isn't nice from architectural point of view, but it's really convenient - compared to some alternatives like pasting it to some file. Any better idea?
 - reusing password for token storage path looks awful, any better idea besides creating new option and passing it down through all the code path?

I've tested it only for caldav, not carddav. And according to http://evertpot.com/google-carddav-issues/ I expect the later to fail badly.

### DIFF
--- a/docs/supported.rst
+++ b/docs/supported.rst
@@ -216,13 +216,15 @@ frequently.
     type = caldav
     url = https://apidata.googleusercontent.com/caldav/v2/
     auth = oauth2_google
-    password = ~/.vdirsyncer/oauth-token
 
     [storage card]
     type = carddav
     url = https://www.googleapis.com/carddav/v1/principals/EMAIL/lists/default
     auth = oauth2_google
-    password = ~/.vdirsyncer/oauth-token
+
+At first run you will be asked to authorize application for google account
+access. Simply follow the instructions. You'll be asked to modify configuration
+file (save `refresh_token` as a password).
 
 - Google's CardDav implementation is very limited, may lead to data loss, use
   with care.
--- a/docs/supported.rst
+++ b/docs/supported.rst
@@ -207,5 +207,26 @@ Vdirsyncer is continuously tested against the latest version of Baikal_.
 Google
 ------
 
-Vdirsyncer doesn't currently support Google accounts fully. For possible
-solutions see :gh:`202` and :gh:`8`.
+Using vdirsyncer with Google Callendar is possible, but it is not tested
+frequently.
+
+::
+
+    [storage cal]
+    type = caldav
+    url = https://apidata.googleusercontent.com/caldav/v2/
+    auth = oauth2_google
+    password = ~/.vdirsyncer/oauth-token
+
+    [storage card]
+    type = carddav
+    url = https://www.googleapis.com/carddav/v1/principals/EMAIL/lists/default
+    auth = oauth2_google
+    password = ~/.vdirsyncer/oauth-token
+
+- Google's CardDav implementation is very limited, may lead to data loss, use
+  with care.
+- You can select which callendars to sync on
+  `CalDav settings page <https://calendar.google.com/calendar/syncselect>`_
+
+For more information see :gh:`202` and :gh:`8`.
--- a/docs/supported.rst
+++ b/docs/supported.rst
@@ -207,7 +207,7 @@ Vdirsyncer is continuously tested against the latest version of Baikal_.
 Google
 ------
 
-Using vdirsyncer with Google Callendar is possible, but it is not tested
+Using vdirsyncer with Google Calendar is possible, but it is not tested
 frequently.
 
 ::
@@ -226,7 +226,7 @@ frequently.
 
 - Google's CardDav implementation is very limited, may lead to data loss, use
   with care.
-- You can select which callendars to sync on
+- You can select which calendars to sync on
   `CalDav settings page <https://calendar.google.com/calendar/syncselect>`_
 
 For more information see :gh:`202` and :gh:`8`.
--- a/vdirsyncer/storage/dav.py
+++ b/vdirsyncer/storage/dav.py
@@ -1,7 +1,15 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import print_function
+
 import datetime
+import errno
+import json
 import logging
+import os
+import sys
+
+from atomicwrites import atomic_write
 
 from lxml import etree
 
@@ -12,8 +20,18 @@ from .base import Item, Storage, normalize_meta_value
 from .http import HTTP_STORAGE_PARAMETERS, USERAGENT, prepare_auth, \
     prepare_client_cert, prepare_verify
 from .. import exceptions, utils
+from ..utils import expand_path
 from ..utils.compat import text_type, to_native
 
+OAUTH2_GOOGLE_TOKEN_URL = 'https://accounts.google.com/o/oauth2/v2/auth'
+OAUTH2_GOOGLE_REFRESH_URL = 'https://www.googleapis.com/oauth2/v4/token'
+have_oauth2 = True
+try:
+    from requests_oauthlib import OAuth2Session
+    oauth2_client_id = 'FIXME'
+    oauth2_client_secret = 'FIXME'
+except ImportError:
+    have_oauth2 = False
 
 dav_logger = logging.getLogger(__name__)
 
@@ -301,7 +319,6 @@ class DavSession(object):
                  useragent=USERAGENT, verify_fingerprint=None,
                  auth_cert=None):
         self._settings = {
-            'auth': prepare_auth(auth, username, password),
             'cert': prepare_client_cert(auth_cert),
         }
         self._settings.update(prepare_verify(verify, verify_fingerprint))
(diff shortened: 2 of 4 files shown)


untitaker:
What a pleasant surprise, thank you! Will review in approx 10 hours

untitaker:
>request and fill google app token (client id, client secret)

I guess that we'll have to make the user set them for now to work around API limits.

>add option to specify them in configuration - I don't know how to get a (global?) config value from inside of Storage object

I'm not happy with having a global option for this. I understand the need for eliminating redundancy, but on the other hand, you're putting stuff into the global section that is only relevant for very few storage types. This seems like a leak in the abstraction to me.

We need a more generic and scalable solution for this anyway (rather than just having new global options), we have the same problem when we access CardDAV and CalDAV of the same server, where credentials and sometimes even URLs are duplicated.

>write some tests? but how to do that without exposing test calendar credentials (you can't simply run local google calendar instance...)? then how to handle interactive google confirmation?

Regarding credentials: Travis can have secret environment variables. We already test a remote DAViCal server like this.

Regarding storing the auth code, I'll have to take a look.

>reusing password for token storage path looks awful, any better idea besides creating new option and passing it down through all the code path?

I'd prefer to have separate storage types for Google CalDAV/CardDAV (probably subclasses of the normal DAV storages, in a separate module), such that we can redefine all parameters.

docs/supported.rst
```diff
@@ -207,5 +207,26 @@ Vdirsyncer is continuously tested against the latest version of Baikal_.
 Google
 ------
 
-Vdirsyncer doesn't currently support Google accounts fully. For possible
-solutions see :gh:`202` and :gh:`8`.
+Using vdirsyncer with Google Callendar is possible, but it is not tested
```
untitaker: *Calendar

docs/supported.rst
```diff
@@ -207,5 +207,26 @@ Vdirsyncer is continuously tested against the latest version of Baikal_.
...
+
+    [storage card]
+    type = carddav
+    url = https://www.googleapis.com/carddav/v1/principals/EMAIL/lists/default
+    auth = oauth2_google
+    password = ~/.vdirsyncer/oauth-token
+
+- Google's CardDav implementation is very limited, may lead to data loss, use
+  with care.
+- You can select which callendars to sync on
```
untitaker: *Calendar

vdirsyncer/storage/dav.py
```diff
@@ -1,7 +1,15 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import print_function
```
hobarrera: Do we need this line? Isn't this just for unsupported python2 versions?

marmarek: On Sat, Mar 26, 2016 at 11:48:43AM -0700, Hugo Osvaldo Barrera wrote:
> Do we need this line? Isn't this just for unsupported python2 versions?

Is python2 really unsupported? I see just a warning about dropping its
support in the future. And also tests are running on python2 (in
addition to python3).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

untitaker:
Also please take a look at `vdirsyncer/storage/remotestorage.py` to keep consistent with UI. Particularly, which functions we use for printing and reading input (and that we open the browser ourselves).

vdirsyncer/storage/dav.py
untitaker: I don't have it in any other file, but don't care.

hobarrera: No, python2 is not unsupported, but IIRC, this line would only be needed for python2 versions we don't support (ie: python26?).

vdirsyncer/storage/dav.py
```diff
@@ -317,6 +349,39 @@ def request(self, method, path, **kwargs):
             url = utils.compat.urlparse.urljoin(self.url, path)
         if self._session is None:
             self._session = requests.session()
+            if self._use_oauth2_google:
```
untitaker: I'd prefer if we have an entirely new storage type for Google access instead of patching that logic in here, but this is a major refactoring effort and I'd be willing to do it myself.

vdirsyncer/storage/dav.py
```diff
@@ -317,6 +349,39 @@ def request(self, method, path, **kwargs):
...
+                    },
+                    token_updater=self._save_oauth2_token
+                )
+                if not self._token:
+                    authorization_url, state = self._session.authorization_url(
+                        OAUTH2_GOOGLE_TOKEN_URL,
+                        # access_type and approval_prompt are Google specific
+                        # extra parameters.
+                        access_type="offline", approval_prompt="force")
+                    print("Please go to %s" % authorization_url)
```
untitaker: Please use `click.echo` here, `print` will behave funky in multithread environments.

See https://github.com/pimutils/vdirsyncer/blob/master/vdirsyncer/storage/remotestorage.py#L96 for similar code.

vdirsyncer/storage/dav.py
```diff
@@ -317,6 +349,39 @@ def request(self, method, path, **kwargs):
...
+                    print("Paste obtained code: ", end='')
+                    code = sys.stdin.readline().strip()
+                    self._token = self._session.fetch_token(
+                        OAUTH2_GOOGLE_REFRESH_URL,
+                        code=code,
+                        # Google specific extra parameter used for client
+                        # authentication
+                        client_secret=oauth2_client_secret,
+                    )
+                    self._save_oauth2_token(self._token)
```
untitaker: Instead of this, I'd prefer if we tell the user to set a new config parameter `access_token` (I guess that's the correct term in OAuth land?). This is a bit more work for the user, but allows them to store the token in a more secure place (e.g. password manager), when combined with https://vdirsyncer.readthedocs.org/en/stable/keyring.html

marmarek:
> I guess that we'll have to make the user set them for now to work around API limits.

Limits looks to be pretty high by default:
- requests/day 	1,000,000 	
- requests/100seconds/user 	500 	

> I'd prefer to have separate storage types for Google CalDAV/CardDAV (probably subclasses of the normal DAV storages, in a separate module), such that we can redefine all parameters.

How to avoid duplication of code in `CaldavStorage` and `CaldavStorage` classes?

Other than that, looks reasonable. That way, it can have those parameters:
- `oauth2_token_storage` - path for storing token
- `oauth2_client_id` - override default value
- `oauth2_client_secret` - override default value

I'll leave that refactoring for you.

> Particularly, which functions we use for printing and reading input (and that we open the browser ourselves).

Is it really good idea to ask user for placing result in the config file? Note that in case of Google, you receive authorization code (not access token), which is one time password to get actual access token and refresh token (used later to request new access tokens).
So that value is used only once and is useless later. But on the other hand, access token is changed frequently (by default expire in one hour) - this is why I've chosen to write it in some file, instead of asking the user for modifying config...

> Regarding credentials: Travis can have secret environment variables. We already test a remote DAViCal server like this.

Ok, will look into that.

Unrelated question: do you want me to squash changes into existing commits (replace them), or create new commits?

untitaker:
>Limits looks to be pretty high by default:

I'll just hardcode some client credentials for now.

>Is it really good idea to ask user for placing result in the config file? Note that in case of Google, you receive authorization code (not access token), which is one time password to get actual access token and refresh token (used later to request new access tokens). So that value is used only once and is useless later. But on the other hand, access token is changed frequently (by default expire in one hour) - this is why I've chosen to write it in some file, instead of asking the user for modifying config...

I guess that means we can ask the user to store the refresh token in the config, and fetch access token each sync?

I really want to avoid complexity here until everything else is stable.

marmarek:
> I guess that means we can ask the user to store the refresh token in the config, and fetch access token each sync?

Ok. But how to exchange authentication code to refresh code then? It needs additional http request ([session.fetch_token](https://github.com/pimutils/vdirsyncer/pull/394/commits/d12ce9e118c324646891cc63edb045fc1da3a44c#diff-74d18da4ca1a200e81bac794b24bfa97R377)). Should it be done by `https://vdirsyncer.5apps.com/`? 
Or as currently in `vdirsyncer`? Then instead of saving it to the file, echo and ask the user for saving it to the config.

> I really want to avoid complexity here until everything else is stable.

I don't see what is complex in `json.dump` and `json.save`... All the complexity is in the protocol, which we can't work around easily :(

untitaker:
>Ok. But how to exchange authentication code to refresh code then? It needs additional http request (session.fetch_token). Should it be done by https://vdirsyncer.5apps.com/?
Or as currently in vdirsyncer? Then instead of saving it to the file, echo and ask the user for saving it to the config.

Yes, that's what I meant:

* Prompt for auth code
* Fetch refresh token
* Tell user to write refresh token

>I don't see what is complex in json.dump and json.save... All the complexity is in the protocol, which we can't work around easily :(

I have a suspicion it will make testing easier since we can just pass the (for testing hardcoded) refresh token as parameter instead of going through all the auth steps.

marmarek:
Added few commits, handled most of above comments.

untitaker:
Great, I'll get to it and make a new PR (for further discussion) when I have the time. Thanks for all your work!

untitaker:
See #399

PR #406, opened 2016-04-06
More bugfixes for Google

untitaker:


terlar:
Yeah, this seems to have fixed it. I now managed to sync all my calendars. All the `ics` files are there and present. At what point does the events get generated / imported? Might be that I have not configured correctly, but right now nothing is put inside the events.

untitaker:
>but right now nothing is put inside the events.

@terlar What do you mean by that? Could you paste such an "empty" file?

terlar:
The events is a directory that gets created, I was planning to use it together with `khal` but it never got popultated. Or perhaps I am understanding things wrong, perhaps khal should go to those folders with the `ics` files.

I think I might have things wrong, it might be that I created that events directory one earlier, and I just point khal to one of these calendars that are pulled down?

terlar:
Yeah, so I configured khal to point to the calendars I was interested in and it seems to work fine. I will try to use it a bit more, but I figure if things got this far it should be all good.

Never mind my first comment.

geier:
After syncing you should have different directories (one for each calendar) which contain `.ics` files (with content that looks something like https://raw.githubusercontent.com/pimutils/khal/master/tests/ics/cal_d.ics ).

untitaker:
Thanks for the response @terlar. Note that on a final release you will probably have to register your own application and change your config accordingly, see #407

terlar:
No problem, thank you for the quick response and fixes, now I am up to speed with vdirsyncer and have started using it. Also thanks for the heads-up, that won't be an issue for me personally as I do the same currently for offlineimap where I have registered my own application.

PR #413, opened 2016-04-07
Fix bug in storage init errorhandler

untitaker:
If there are no defaults, spec.args would be sliced like
`spec.args[1:0]` which always produces an empty slice.


untitaker:
@homu r+

Issue #416, opened 2016-04-07
Google error: 405 Method not allowed for url


IronWolfMC:
vdirsyncer version: 0.9.4.dev35+ng777eb35
Python: 3.4.3
OS: Ubuntu 14.04 Server (Docker)

Google returns a 405 Client Error (Method Not Allowed for url: https://apidata.googleusercontent.com/caldav/v2/) when attempting to synchronize with a Google calendar. Both the Google Calendar API and the Caldav API are enabled.

Config:
```
[general]
status_path = ~/.config/vdirsyncer/status/

[pair my_calendars]
a = my_cal
b = my_gmail
collections = null

[storage my_cal]
type = caldav
url = http://my.owncloud.server/cloud/remote.php/caldav/calendars/USERNAME/personal
username = USERNAME
password = ****************

[storage my_gmail]
type = google_calendar
token_file = ~/.google-token
client_id = ************************************************.apps.googleusercontent.com
client_secret = ********************
```

Error: (Gist: https://gist.github.com/IronWolfMC/654a22cc583d09322b5d6d183a72371c)
```
root@bbeb76ec35a4:~# vdirsyncer sync
Syncing my_calendars
warning: Skipping '/caldav/v2/', etag property is missing.
Unhandled exception occured for my_calendars.
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/vdirsyncer-0.9.4.dev35+ng777eb35-py3.4.egg/vdirsyncer/cli/tasks.py", line 66, in sync_collection
    force_delete=force_delete
  File "/usr/local/lib/python3.4/dist-packages/vdirsyncer-0.9.4.dev35+ng777eb35-py3.4.egg/vdirsyncer/sync.py", line 216, in sync
    b_info.prepare_idents()
  File "/usr/local/lib/python3.4/dist-packages/vdirsyncer-0.9.4.dev35+ng777eb35-py3.4.egg/vdirsyncer/sync.py", line 124, in prepare_idents
    if prefetch else ()):
  File "/usr/local/lib/python3.4/dist-packages/vdirsyncer-0.9.4.dev35+ng777eb35-py3.4.egg/vdirsyncer/storage/dav.py", line 423, in get_multi
    headers=self.session.get_default_headers()
  File "/usr/local/lib/python3.4/dist-packages/vdirsyncer-0.9.4.dev35+ng777eb35-py3.4.egg/vdirsyncer/storage/dav.py", line 328, in request
[...]

PR #418, opened 2016-04-08
Ability to use other collection for missing one

untitaker:
Vdirsyncer asks you if it should create a new collection if the current one is missing, but why not allow using another collection in place?

This idea still needs a lot of UI work. It also shouldn't be possible to "reuse" a collection more than once. Given the amount of complexity a properly polished implementation adds, I'm not sure it's a good idea anyway.

### DIFF
--- a/vdirsyncer/cli/utils.py
+++ b/vdirsyncer/cli/utils.py
@@ -286,9 +286,10 @@ def _handle_collection_not_found(config, discovered, collection, e=None):
                 if display_name else ''
             ))
 
-        collection_i = click.prompt('Your choice?',
-                                  type=click.IntRange(min=0,
-                                                      max=len(choices)))
+        collection_i = click.prompt(
+            'Your choice?',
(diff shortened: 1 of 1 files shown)


untitaker:
(cont #416)

@IronWolfMC Check this out, using the second config you tried (the original config with the suggestion from https://github.com/pimutils/vdirsyncer/issues/416#issuecomment-207203700).

untitaker:
Also given that you can just symlink everything together, I am *really* not sure if this change matters.

marmarek:
In case of one backend being a file, you can symlink. But if you for
example sync one caldav with another one (say, google), it isn't
possible (without introducing another complexity of intermediate
storage)..

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

untitaker:
https://github.com/pimutils/vdirsyncer/pull/423 is the successor.

PR #421, opened 2016-04-09
Use displayname for collection discovery

### LINKED ISSUE #420: Mapping of collections is unflexible
As demonstrated in #416 and mentioned in #418, the current way of mapping collections is unflexible. In the most common usecase (`DAV <=> filesystem`) this only means that collection's folder names are determined by the server, which can be easily worked around with symlinks, but when syncing server-to-server, you might be unable to correctly map already existing collections against each other.

This is related to #405, but not the same issue.

There are multiple ways to solve this:

* Have a interactive UI to connect collections from each side with each other when running `vdirsyncer discover`. A primitive implementation of this is #418.
* Somehow extend the configuration format for this. Unsure how this would look like.
* Simply match by displayname (easiest). This requires:
  * That the user has unique displaynames set for each calendar per-side (this should be rigorously checked)
  * That all servers support displaynames
  * That every collection has a non-empty displayname

untitaker:
Fix #420

I'd really like to get some input before this gets landed in master, because this change is quite the breakage.

### DIFF
--- a/vdirsyncer/cli/utils.py
+++ b/vdirsyncer/cli/utils.py
@@ -221,6 +221,7 @@ def _expand_collections_cache(collections, config_a, config_b):
 
 def _discover_from_config(config):
     storage_type = config['type']
+    instance_name = config.get('instance_name', '???')
     cls, config = storage_class_from_config(config)
 
     try:
@@ -230,7 +231,7 @@ def _discover_from_config(config):
             raise exceptions.UserError(
                 'The storage {} (type {}) doesn\'t support collection '
                 'discovery. You can only use `collections = null` with it.'
-                .format(config.get('instance_name', '???'), storage_type)
+                .format(instance_name, storage_type)
             )
     except Exception:
         return handle_storage_init_error(cls, config)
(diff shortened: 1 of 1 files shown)


untitaker:
This has the strong assumption that the user is able and willing to rename collections on each side to the same name.

Another option I have considered is a option flag (default to true), but I don't think it's worth the complexity.

Unfortunately we get a lot of feedback only after a release, so we might just have to see pre-1.0.

marmarek:
Not sure if the best option (if enabled unconditionally). For example I have some calendar (google) for which I can't change the name, and that name is not much descriptive for me. So I would rather avoid using the same name on my local (owncloud) copy.
Details: my contract provider have calendar for each contractor, named with contractor name. From his point of view this makes a perfect sense, but for me it's just another calendar with the same name...

Also in such a case, if that calendar would be renamed, the other one needs to be renamed too. It may be very inconvenient especially when one doesn't have control over one of those calendars.

I think if would be good idea as a configurable feature, but not enabled unconditionally. #418 is in good direction.

BTW for google calendars there is no problem with getting unique calendar ID, as calendar IDs are unique.

untitaker:
@marmarek What if we merge this, but also allow configurations like this:

    collections = [
        ["config_name1", "remote_name1", "local_name1"],
        "name2",
        "name3"
    ]


* `"config_name1" would be used as argument for `vdirsyncer repair`, `vdirsyncer sync` and in its output.

* `"remote_name1" would be used for side A

* `"local_name1" would be used for side B

untitaker:
I think I have a better idea. Will take a while

marmarek:
That would be great!
But I think this is only slightly related to this PR. I see two things:
 - what should be used as collection name (this PR)
 - how to match collections with each other (your idea)

When one doesn't control one of remote collections, we can't assume for example its being unique...

untitaker:
>When one doesn't control one of remote collections, we can't assume for example its being unique...

That's true. I fear though that

untitaker:
Hit enter too early, edited previous comment.

marmarek:
As noted earlier - at least for google, calendar ID (next to last URL element) is a stable ID. Not sure how to proceed with "plain caldav".

untitaker:
OK, closing for real this time.

PR #431, opened 2016-04-18
Add pypy3 to test matrix

### LINKED ISSUE #429: Test against pypy Python 3 in CI
I want vdirsyncer to be able to run on some version of Pypy. If we're going to kill support for Python 2, we should activate the pypy3 Python version in Travis.

untitaker:
Fix #429

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -167,6 +167,18 @@
         "python": "pypy"
       },
       {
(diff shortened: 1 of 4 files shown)


untitaker:
@homu r+

untitaker:
This is not feasible at the moment, too many libraries are broken on pypy3.

PR #434, opened 2016-04-22
Replace LXML with stdlib XML

### LINKED ISSUE #433: Switch from lxml to stdlib xml
lxml is very slow to build, and even though it might be a bit faster, the major performance problems are in vdirsyncer's code anyway.

@geier You built khal's CalDAV support with lxml. Was there any particular reason?

untitaker:
Just because the XML module is already there, and suits our needs as
well. DoS concerns should be irrelevant as we trust the server to that
extent.

Fix #433


untitaker:
@homu r+

Issue #430, opened 2016-04-17
macos launchd .plist


benjaminfrank:
I created a vdirsyncer.plist (somewhat like a cronjob) to have vdirsyncer run every minute. would something like this qualify for inclusion into mainline? I am asking because of https://github.com/Homebrew/homebrew-core/pull/348
maybe the default values should be a bit more conservative as well...

benjaminfrank:
I can easily ceate a PR. where would I put such a crontrib script? make a new contrib folder or something?

hobarrera:
What's the difference between a plain-old crontab entry and a plist file? (not an OS X user, so I simply don't know).

benjaminfrank:
crontab is deprecated on mac and its functionality included in launchd (the thing that runs those .plist files). launchd also autostarts & keeps daemons alive and all this fancy stuff

untitaker:
What happens if vdirsyncer itself takes longer than a minute? Is a second job spawned?

In principle I wouldn't run two sync processes on the same collection, although the race condition created by it seems to be extremely unlikely.

In principle I'd happily merge configuration files that could be reused between distros (systemd files come to mind), but I'm somewhat reluctant about cronjobs because of the uncertainity above. I'd much rather solve #24 but that one has a long way to go.

benjaminfrank:
Yes, a second one is run. Maybe launchd has some facility to prevent this, I need to check.
On vdirsyncer side a pid file could alleviate this issue, although #24 sounds good as well.
maybe I'll check into the pid file thing, that should be reasonably easy to implement

untitaker:
Ideally one would have one PID file per collection, stored in the status directory in a similar scheme to .items files

On 18 April 2016 08:46:52 CEST, benjaminfrank <notifications@github.com> wrote:
>Yes, a second one is run. Maybe launchd has some facility to prevent
>this, I need to check.
>On vdirsyncer side a pid file could alleviate this issue, although #24
>sounds good as well.
>maybe I'll check into the pid file thing, that should be reasonably
>easy to implement
>
>---
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/pimutils/vdirsyncer/issues/430#issuecomment-211229427

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

benjaminfrank:
[...]

PR #435, opened 2016-04-23
Initial MacOS X launchd .plist file.

benjaminfrank:
PR for #430

### DIFF
--- a/contrib/vdirsyncer.plist
+++ b/contrib/vdirsyncer.plist
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!-- Blueprint for cron-like launchd plist -->
+<!-- Replace @@PLACEHOLDERS@@ with appropriate values for your system/settings! -->
+<plist version="1.0">
+  <dict>
+    <key>EnvironmentVariables</key>
+    <dict>
+      <!-- Locale to use for vdirsyncer, e.g. en_US.UTF-8 -->
+      <key>LANG</key>
+      <string>@@LOCALE@@</string>
+      <key>LC_ALL</key>
+      <string>@@LOCALE@@</string>
+    </dict>
+    <key>Label</key>
+    <string>vdirsyncer</string>
+    <key>WorkingDirectory</key>
+    <!-- working directory for vdirsyncer, usually the base directory where
+         vdirsyncer is installed, e.g. /usr/local/ -->
+    <string>@@WORKINGDIRECTORY@@</string>
+    <key>ProgramArguments</key>
+    <array>
+      <!-- full path to vdirsyncer binary -->
+      <string>@@VDIRSYNCER@@</string>
+      <!-- only log errors -->
+      <string>-v</string>
+      <string>ERROR</string>
+      <string>sync</string>
+    </array>
+    <key>RunAtLoad</key>
+    <true/>
+    <key>StartInterval</key>
+    <!-- Sync intervall in seconds -->
+    <integer>@@SYNCINTERVALL@@</integer>
+    <!-- For logging, redirect stdout & stderr -->
+    <!-- <key>StandardErrorPath</key> -->
+    <!-- Full path to stderr logfile, e.g. /tmp/vdirsyncer_err.log -->
+    <!-- <string>@@STDERRFILE@@</string>  -->
+    <!-- Full path to stdout logfile, e.g. /tmp/vdirsyncer_out.log -->
+    <!-- <key>StandardOutPath</key> -->
+    <!-- <string>@@STDOUTFILE@@</string> -->
+  </dict>
+</plist>
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,6 +8,7 @@ include docs-requirements.txt
 
 recursive-include docs *
 recursive-include tests *
+recursive-include contrib *
 prune tests/storage/servers
 recursive-include tests/storage/servers/radicale *
 recursive-include tests/storage/servers/skip *


contrib/vdirsyncer.plist
```diff
@@ -0,0 +1,34 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <!-- click requires a proper UTF-8 locale -->
+    <key>EnvironmentVariables</key>
+    <dict>
+           <key>LANG</key>
+           <string>en_US.UTF-8</string>
+           <key>LC_ALL</key>
+           <string>en_US.UTF-8</string>
```
untitaker: So there's generally a lot of confusion around this: Click generally *respects* the locale, whatever it is set to. Python 3 implies that a missing locale configuration implies the `C` locale -- Click treats both conditions as an error, since it's more likely a result of a misconfigured environment (with missing envvars) rather than intention. Unfortunately it can't detect if the C locale is intentionally set, but usually it isn't.

A few questions:

* What is `LANG`/`LC_ALL` set by default in launchd "cronjobs"?
* What are other plists doing?

benjaminfrank: I don't know what exactly the default is, but vdirsyncer (or rather click) bails out with this UTF-8 error (which I can't remember right now, but it tells you to properly set your LANG & LC_ALL).
Launchd does not use you .profile or anything, I wouldn't even know which shell they use internally or if the use a proper shell at all or rather fork() or something.

In launchd you can set env vars by calling ```launchctl setenv <VAR> <VALUE>``` however this is not persisted. The only way to "persist" such a setting is to create a .plist that sets those values after booting. this has (at least) 2 disadvantages: a) execution order can't be controlled in launchd b) it creates a dependency on another plist

So I figured setting those vars inside the plist is the most hasslefree way of doing things.
Other plist are usually some startup scripts rather than cron like jobs, e.g. in homebrew there is no cron like plist.

The more I think about it, this plist takes quite a few assumptions, i.e. where vdirsyncer is installed. Maybe it would make more sense to put some comment ala "this is an example, please adjust as needed"o n top? For the homebrew one, I need to change this anyway so it properly adjust to where homebrew install things etc.

untitaker: I have no problem with including assumptions (even homebrew-specific ones) in the plist because users understand how to change them, but using the wrong locale may break things in a very subtle way on POSIX (corrupted filenames etc), if theoretically your real locale is a very obscure one. I don't know if such a locale is even possible on OS X. But either way, it is technically wrong to assume any hardcoded value.

I figured there must be a way to get the proper locale variables, maybe as an extra command before vdirsyncer. I think at this point we would need some third-party opinion/knowledge.

On 24 April 2016 13:09:59 CEST, benjaminfrank <notifications@github.com> wrote:
>> @@ -0,0 +1,34 @@
>> +<?xml version="1.0" encoding="UTF-8"?>
>> +<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
>"http://www.apple.com/DTDs/PropertyList-1.0.dtd">
>> +<plist version="1.0">
>> +  <dict>
>> +    
>> +    <key>EnvironmentVariables</key>
>> +    <dict>
>> +           <key>LANG</key>
>> +           <string>en_US.UTF-8</string>
>> +           <key>LC_ALL</key>
>> +           <string>en_US.UTF-8</string>
>
>I don't know what exactly the default is, but vdirsyncer (or rather
>click) bails out with this UTF-8 error (which I can't remember right
>now, but it tells you to properly set your LANG & LC_ALL).
>Launchd does not use you .profile or anything, I wouldn't even know
>which shell they use internally or if the use a proper shell at all or
>rather fork() or something.
>
>In launchd you can set env vars by calling ```launchctl setenv <VAR>
><VALUE>``` however this is not persisted. The only way to "persist"
>such a setting is to create a .plist that sets those values after
>booting. this has (at least) 2 disadvantages: a) execution order can't
>be controlled in launchd b) it creates a dependency on another plist
>
>So I figured setting those vars inside the plist is the most hasslefree
>way of doing things.
>Other plist are usually some startup scripts rather than cron like
>jobs, e.g. in homebrew there is no cron like plist.
>
>The more I think about it, this plist takes quite a few assumptions,
>i.e. where vdirsyncer is installed. Maybe it would make more sense to
>put some comment ala "this is an example, please adjust as needed"o n
>top? For the homebrew one, I need to change this anyway so it properly
>adjust to where homebrew install things etc.
>
>---
>You are receiving this because you commented.
>https://github.com/pimutils/vdirsyncer/pull/435/files/89181d1cb25be28a22d3fe436ab3b5b79c1487b0#r60842852

-- 

benjaminfrank: The people over @homebrew suggest the best way would be to have some placeholder values like @@PREFIX@@ for such values. rationale is that people either use some sort of package manager, which can replace those values with the actual ones, or install manually where they need to provide them anyway. not really sure how this would work e.g. with pip.
but for homebrew this would be the best solution and give vdirsyncer user a blueprint in case they install manually. this would also alleviate your concern regarding hard-coded values. would that be ok?

untitaker: I now did a bit of research on this and indeed there doesn't seem to be a way to do this the way I imagined (dynamically setting the correct locale at runtime). I don't really care about the value itself, but it should be made clear to the reader that this part may have to be changed to the correct locale string.

untitaker:
This looks good, do you think it should be included in the tarballs on PyPI?

contrib/vdirsyncer.plist
benjaminfrank: can you share a bit what the result of your research is? while my locale is de_DE.UTF-8 I am using en_US.UTF-8 and had no issues so far. But I wonder what issues one could encounter?

benjaminfrank:
Sure, but I'd like to alter the PR to use this placeholder syntax first, so the resulting .plist needs to be adjusted by the user (or the package manager, as I intend to do for homebrew)

untitaker:
Sure, I have no problem with that. Please also include one or two sentences in a comment to explain the variable.

contrib/vdirsyncer.plist
untitaker: I only focused on how to fetch the system-wide locale and set the envvars accordingly. `de_DE.UTF-8` and `en_US.UTF-8` are the same as far as vdirsyncer is concerned. The situation gets really nasty if you mix some `ISO-8859-6`-encodings/locales with UTF-8 ones:

    >>> 'ä'.encode('utf-8').decode('ISO-8859-6')
    'أ¤'

benjaminfrank:
Updated the placeholders & comments. Please have a quick look if this makes sense to you.

untitaker:
Very nice. Curious, do these placeholders get picked up by homebrew?

untitaker:
Also please add `recursive-include contrib *` to MANIFEST.in to include the files into the PyPI tarballs.

benjaminfrank:
Homebrew does not pick those placeholders up automatically, but it has facilities to get those values (except the locale I think, that's why I was curious about the research you mentioned) and allows inline replacement. together this allows a fairly easy and programmatic replacement of such placeholders.

untitaker:
Thank you!

contrib/vdirsyncer.plist
benjaminfrank: So basically a user needs to set the locale to the one used for creating the entries/the server uses?

untitaker: No, to the locale of the system vdirsyncer runs on.

On 26 April 2016 09:28:34 CEST, benjaminfrank <notifications@github.com> wrote:
>> @@ -0,0 +1,34 @@
>> +<?xml version="1.0" encoding="UTF-8"?>
>> +<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
>"http://www.apple.com/DTDs/PropertyList-1.0.dtd">
>> +<plist version="1.0">
>> +  <dict>
>> +    
>> +    <key>EnvironmentVariables</key>
>> +    <dict>
>> +           <key>LANG</key>
>> +           <string>en_US.UTF-8</string>
>> +           <key>LC_ALL</key>
>> +           <string>en_US.UTF-8</string>
>
>So basically a user needs to set the locale to the one used for
>creating the entries/the server uses?
>
>---
>You are receiving this because you modified the open/close state.
>https://github.com/pimutils/vdirsyncer/pull/435/files/89181d1cb25be28a22d3fe436ab3b5b79c1487b0#r61039109

-- 

Issue #404, opened 2016-04-05
Authentication of two Google storages at once will mix lines


untitaker:
When syncing both Google calendar and contacts for the first time, it will open both auth urls at the same time, and also prompt for both codes serially.

This needs changes in click-threading.

marmarek:
What about adding documentation like "when syncing google for the first time, sync one pair at the time"?

untitaker:
I guess that's the workaround, but I think it is easy enough to fix this properly.

untitaker:
I'm unsure about the API I want to expose in click-threading. I think I'll have to properly document it before fixing this issue, so others can help :(

untitaker:
Related: https://github.com/click-contrib/click-threading/issues/3

PR #437, opened 2016-04-26
Refactor UX of discovery

### LINKED ISSUE #419: Printed result of `vdirsyncer discover` contains duplicate entries
If all collections exist on both sides, and `collections = ["from a", "from b"]` is set in the config, the output looks like this:

    $ vdirsyncer discover dav_calendars
    Discovering collections for pair dav_calendars
    Saved for dav_calendars: collections = ["ssl-certs", "holidays", "default", "media", "computers", "holidays-auto", "ssl-certs", "default", "holidays", "media", "computers", "holidays-auto"]

Internally there is no such duplication, only the output is affected.

### LINKED ISSUE #424: Collection names have to be guessed by the user
The concept of a "collection name" is not exposed through UIs outside of vdirsyncer. Consequently the user doesn't know a thing about them, and when wanting to explicitly define them (i.e. not just use `["from a", "from b"]`), they are left guessing.

#421 was an attempt at fixing this, but I think we could do better:

## Select by displayname

    [pair foo]
    # both mean the same thing
    collections = ["displayname/Marco's work calendar"]
    collections = ["work"]
   
## Have a command for listing collections

`vdirsyncer discover --list foo` would list all collection names including displayname. Then you can just copypaste them into the config

untitaker:
- The new `--list` option for `vdirsyncer discover`, enabled by default,
  makes vdirsyncer list all collections before saving the configuration.
  It is not enabled when vdirsyncer tries to discover when invoked via
  `vdirsyncer sync`. Fix #424

- There are no duplicate collections anymore. Fix #419.

- `vdirsyncer discover` is running with one worker by default now, to
  avoid broken output. See also #404.

### DIFF
--- a/vdirsyncer/cli/__init__.py
+++ b/vdirsyncer/cli/__init__.py
@@ -74,7 +74,6 @@ def max_workers_callback(ctx, param, value):
     return value
 
 
-
 def max_workers_option(default=0):
     help = 'Use at most this many connections. '
     if default == 0:
--- a/vdirsyncer/cli/__init__.py
+++ b/vdirsyncer/cli/__init__.py
@@ -74,13 +74,20 @@ def max_workers_callback(ctx, param, value):
     return value
 
 
-max_workers_option = click.option(
-    '--max-workers', default=0, type=click.IntRange(min=0, max=None),
-    callback=max_workers_callback,
-    help=('Use at most this many connections. With debug messages enabled, '
-          'the default is 1, otherwise one connection per collection is '
-          'opened.')
-)
+
+def max_workers_option(default=0):
+    help = 'Use at most this many connections. '
+    if default == 0:
+        help += 'The default is 0, which means "as many as necessary". ' \
+                'With -vdebug enabled, the default is 1.'
+    else:
+        help += 'The default is {}.'.format(default)
+
+    return click.option(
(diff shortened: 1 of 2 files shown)


untitaker:
Open question: Should `vdirsyncer sync` attempt automatic discover (the current state), or should this process be triggered by the user, since it's becoming increasingly interactive?

hobarrera:
I think everything is simpler (code-wise) if we require it to be run manually - and it also makes sense for users to configure vdirsyncer *once* before blindly making it sync.

In other words, both options have pros from a user perspective, but for us, no-automatic discovery seems simpler to implement/maintain/support.

untitaker:
I'd like to ignore the maintainability aspect completely in this decision. Discovery has been such a huge pain point for server-to-server sync that it's not even funny anymore.

marmarek:
I'm using vdirsyncer for server-to-server case. I really like the idea of
manual discovery call first.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

untitaker:
I've just noticed that it's not server-to-server sync in particular: It's a problem when you're trying to connect two storages where all collections already exist on both sides, and vdirsyncer is unable to match them.

PR #464, opened 2016-06-09
Improve google documentation

evgeni:


vdirsyncer/storage/google.py
```diff
@@ -94,6 +94,11 @@ def _save_token(token):
 class GoogleCalendarStorage(dav.CaldavStorage):
     __doc__ = '''Google calendar.
 
+    .. note:: You have to poperly set up ``collections``, otherwise errors like
```
untitaker: Not sure if that's useful. `collections` always needs to be properly set up?

untitaker: At this point I'm actually curious which problem led you to write this. Stuff like #463? Though that typo should be properly caught when parsing the config...

evgeni: The example in "tutorial" (for ownClould) has `collections = null` and "config" does not list that as a required field. Thus I thought one only has to set `collections` when syncing non-default adressbooks/calendars.

untitaker: Ah, so I should fix the google storage to reject `collections = null`.

evgeni: If that's feasible, yeah.
Should I drop the changes to `google.py` from that PR then?

untitaker: I think https://github.com/pimutils/vdirsyncer/pull/466 would fix this, so you can take this paragraph out.

untitaker: Sorry, didn't get a push notification about your reply... yes, please drop those changes.

evgeni:
@untitaker done :)

untitaker:
Thanks!

PR #488, opened 2016-08-14
Introduce collections for singlefile storage

### LINKED ISSUE #485: Singlefile storage should support collections
As discussed in #479, `singlefile`-storage should create separate files for each collection. I'm thinking of something like:

    [storage singlefile]
    path = /foo/bar/%s.ics

and `%s` will be replaced by the collection name.

untitaker:
Fix #485

### DIFF
--- a/vdirsyncer/storage/singlefile.py
+++ b/vdirsyncer/storage/singlefile.py
@@ -3,6 +3,7 @@
 import collections
 import contextlib
 import functools
+import glob
 import logging
 import os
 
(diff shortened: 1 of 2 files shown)


vdirsyncer/storage/singlefile.py
```diff
@@ -40,14 +41,33 @@ class SingleFileStorage(Storage):
         This storage is very slow, and that is unlikely to change. You should
         consider using :storage:`filesystem` if it fits your usecase.
 
-    :param path: The filepath to the file to be written to.
+    :param path: The filepath to the file to be written to. If collections are
+        used, this should contain ``%s`` as a placeholder for the collection
+        name.
     :param encoding: Which encoding the file should use. Defaults to UTF-8.
 
     Example for syncing with :storage:`caldav`::
```
untitaker: Hello @gour, could you try this out? The example for using collections with singlefile storage starts here.

PR #499, opened 2016-09-07
Remove Python 2 support

### LINKED ISSUE #219: Discontinue Python 2 support
## urlencoding

I used to think that Python 3 will be more daunting in Unicode support, but
I've been wrong: Both Python 2 and 3 have usually been equally hard to support.

Turns out that this was only the case until #218 and #215: urllib and urlparse
are completely broken under Python 2, they don't support unicode strings at
all:

    $ python2
    Python 2.7.10 (default, May 26 2015, 04:16:29) 
    [GCC 5.1.0] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    >>> from urllib import quote
    >>> quote(u'lolä')
    /usr/lib/python2.7/urllib.py:1303: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode
     - interpreting them as being unequal
      return ''.join(map(quoter, s))
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python2.7/urllib.py", line 1303, in quote
        return ''.join(map(quoter, s))
    KeyError: u'\xe4'
    >>> quote('lolä')
    'lol%C3%A4'
    >>> 

Supporting URLs with unicode chars in them was (and is) a major pain in the
ass, on both versions. The stdlib on Python 2 doesn't help.

I'm not going to say that Python 3 is a strictly superior language: In fact I
think it will introduce severe performance regressions, and e.g. the string
model it uses for Unix filesystem APIs is oversimplifying things simply for the
sake of simplicity, not because that makes it more correct. This is wrong in
theory, in practice it would make things easier for me if I didn't have to
support Python 2.

~~I don't want to drop support for Python 2 in the next years, but perhaps stop
fixing unicode bugs on it.~~ FUCK IT

## other things

* threading and `^C`
* Semicolons in config values (found in #317): https://bugs.python.org/issue16374
* Duplicate sections are ignored in configs, see #163 

---

## plan

- [x] show warning... 0.8
- [x] require confirmation through config option... 0.11
- [ ] just crash, remove config option

---

fixes #315 
fixes #316 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21973151-discontinue-python-2-support?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github).
</bountysource-plugin>

untitaker:
Fix #219

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,70 +16,6 @@
     "include": [
       {
         "env": "BUILD=style BUILD_PRS=true",
-        "python": "2.7"
-      },
-      {
-        "env": "BUILD=test REMOTESTORAGE_SERVER=mysteryshack REQUIREMENTS=devel BUILD_PRS=false",
-        "python": "2.7"
-      },
-      {
-        "env": "BUILD=test REMOTESTORAGE_SERVER=mysteryshack REQUIREMENTS=release BUILD_PRS=false",
-        "python": "2.7"
-      },
-      {
-        "env": "BUILD=test REMOTESTORAGE_SERVER=mysteryshack REQUIREMENTS=minimal BUILD_PRS=false",
-        "python": "2.7"
-      },
-      {
-        "env": "BUILD=test DAV_SERVER=owncloud REQUIREMENTS=devel BUILD_PRS=false",
-        "python": "2.7"
-      },
-      {
-        "env": "BUILD=test DAV_SERVER=owncloud REQUIREMENTS=release BUILD_PRS=false",
-        "python": "2.7"
-      },
-      {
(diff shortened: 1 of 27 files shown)


hobarrera:
I assume that `compat.str`, `compat.to_native`, etc are all pending a refactor, right?

untitaker:
I'm doing this step-by-step and want to have as small as possible commits.

PR #500, opened 2016-09-13
Add ldap storage

### LINKED ISSUE #283: LDAP support
Hi @all,

I wonder wether it would be a good idea to implement LDAP support as a storage strategy.
Actually I am curious into implementing such code myself, as I need it myself... ;-)

Cheers,
poikilotherm

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27631945-ldap-support?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F574345&utm_medium=issues&utm_source=github).
</bountysource-plugin>

paulfariello:
Current implementation is read only.
Not sure if ldap storage is supposed to do vcard conversion.

I haven't added tests nor documentation but will if @pimutils thinks this patch is meaningful.

Other enhancement would be to allow for attributes configuration.

---

edit by @untitaker: Fix #283

### DIFF
--- a/vdirsyncer/storage/ldap.py
+++ b/vdirsyncer/storage/ldap.py
@@ -53,7 +53,12 @@ class LDAPStorage(Storage):
         for entry in self.conn.entries:
             ldap_logger.debug('Found {}'.format(entry.entry_get_dn()))
             href = entry.entry_get_dn()
-            etag = str(entry.whenChanged)
+            if getattr(entry, 'whenChanged'):
+                etag = str(entry.whenChanged)
+            else:
+                entry = self.get(href)
+                etag = item.hash
+
             yield href, etag
 
     def get(self, href):
--- a/vdirsyncer/storage/ldap.py
+++ b/vdirsyncer/storage/ldap.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 import ldap3
+import vobject
 
 from .base import Item, Storage
 from .. import exceptions
@@ -23,7 +24,9 @@ class LDAPStorage(Storage):
     item_mimetype = 'text/vcard'
 
     def __init__(self, uri=None, search_base=None, bind=None, password=None,
-                 filter='(objectClass=*)', **kwargs):
+                 filter='(&(objectCategory=person)(objectClass=user)'
+                        '(sn=*)(givenName=*))',
+                 **kwargs):
         super(LDAPStorage, self).__init__(**kwargs)
         self.search_base = search_base
         self.filter = filter
@@ -55,7 +58,8 @@ class LDAPStorage(Storage):
         self.conn.search(href, self.filter,
                          attributes=["whenChanged", "cn", "sn", "givenName",
                                      "displayName", "telephoneNumber",
-                                     "mobile", "mail"])
+                                     "mobile", "facsimileTelephoneNumber",
+                                     "mail", "title"])
 
         if not self.conn.entries[0]:
             raise exceptions.NotFoundError(href)
@@ -63,22 +67,32 @@ class LDAPStorage(Storage):
         entry = self.conn.entries[0]
         etag = str(entry.whenChanged)
 
-        vcard = "BEGIN:VCARD\r\n"
-        vcard += "VERSION:3.0\r\n"
-        vcard += "FN;CHARSET=UTF-8:{cn}\r\n".format(cn=entry.cn)
-        if getattr(entry, 'sn', None):
-            vcard += "N;CHARSET=UTF-8:{sn};{givenName}\r\n".format(
-                givenName=entry.givenName, sn=entry.sn)
+        vcard = vobject.vCard()
+        vo = vcard.add('fn')
+        vo.value = str(entry.cn)
+        vo = vcard.add('n')
+        vo.value = vobject.vcard.Name(family=str(entry.sn),
+                                      given=str(entry.givenName))
         if getattr(entry, 'telephoneNumber', None):
-            vcard += "TEL;WORK;VOICE:{tel}\r\n".format(
-                tel=entry.telephoneNumber)
+            vo = vcard.add('tel')
+            vo.value = str(entry.telephoneNumber)
+            vo.type_param = 'WORK'
         if getattr(entry, 'mobile', None):
-            vcard += "TEL;CELL;VOICE:{mobile}\r\n".format(mobile=entry.mobile)
+            vo = vcard.add('tel')
+            vo.value = str(entry.mobile)
+            vo.type_param = 'CELL'
+        if getattr(entry, 'facsimileTelephoneNumber', None):
+            vo = vcard.add('tel')
+            vo.value = str(entry.facsimileTelephoneNumber)
+            vo.type_param = 'FAX'
         if getattr(entry, 'mail', None):
-            vcard += "EMAIL;INTERNET:{email}\r\n".format(
-                email=entry.mail.value.strip())
-        vcard += "END:VCARD"
+            vo = vcard.add('email')
+            vo.value = str(entry.mail)
+            vo.type_param = 'INTERNET'
+        if getattr(entry, 'title', None):
+            vo = vcard.add('title')
+            vo.value = str(entry.title)
 
-        item = Item(vcard)
+        item = Item(vcard.serialize())
 
         return item, etag
--- a/vdirsyncer/storage/ldap.py
+++ b/vdirsyncer/storage/ldap.py
@@ -60,7 +60,7 @@ class LDAPStorage(Storage):
             if getattr(entry, 'whenChanged'):
                 etag = str(entry.whenChanged)
             else:
-                entry = self.get(href)
+                item = self.get(href)
                 etag = item.hash
 
             yield href, etag
@@ -110,7 +110,8 @@ class LDAPStorage(Storage):
 
     def upload(self, item):
         vcard = vobject.readOne(item.raw)
-        self.conn.strategy.add_entry('cn={},ou=test,o=lab'.format(vcard.fn), vcard)
+        self.conn.strategy.add_entry('cn={},ou=test,o=lab'.format(vcard.fn),
+                                     vcard)
 
     def update(self, href, item, etag):
         vcard = vobject.readOne(item.raw)
--- a/vdirsyncer/storage/ldap.py
+++ b/vdirsyncer/storage/ldap.py
@@ -21,15 +21,15 @@ class LDAPStorage(Storage):
     fileext = '.vcf'
     item_mimetype = 'text/vcard'
 
-    def __init__(self, url='ldap://localhost', search_base=None, bind=None, password=None,
+    def __init__(self, url='ldap://localhost', search_base=None, bind=None,
+                 password=None,
                  filter='(&(objectCategory=person)(objectClass=user)'
                         '(sn=*)(givenName=*))',
-                 conn=None,
-                 **kwargs):
+                 _conn=None, **kwargs):
         super(LDAPStorage, self).__init__(**kwargs)
         self.search_base = search_base
         self.filter = filter
-        self.conn = conn
+        self.conn = _conn
         if self.conn is None:
             server = ldap3.Server(url, get_info=ldap3.DSA)
             if bind:
@@ -37,8 +37,9 @@ class LDAPStorage(Storage):
                                              password=password)
             else:
                 self.conn = ldap3.Connection(server)
-        self.conn.bind()
-        self.conn.start_tls()
+            self.conn.bind()
+            self.conn.start_tls()
+
         ldap_logger.debug('Connected to: {}'.format(self.conn))
 
         if self.search_base is None:
--- a/vdirsyncer/storage/ldap.py
+++ b/vdirsyncer/storage/ldap.py
@@ -11,7 +11,7 @@ ldap_logger = logging.getLogger(__name__)
 
 class LDAPStorage(Storage):
     '''
-    :param uri: LDAP URI
+    :param url: LDAP URL
     :param search_base: search base
     :param bind: bind dn
     :param password: bind password
@@ -22,7 +22,7 @@ class LDAPStorage(Storage):
     fileext = '.vcf'
     item_mimetype = 'text/vcard'
 
-    def __init__(self, uri='ldap://localhost', search_base=None, bind=None, password=None,
+    def __init__(self, url='ldap://localhost', search_base=None, bind=None, password=None,
                  filter='(&(objectCategory=person)(objectClass=user)'
                         '(sn=*)(givenName=*))',
                  conn=None,
@@ -32,7 +32,7 @@ class LDAPStorage(Storage):
         self.filter = filter
         self.conn = conn
         if self.conn is None:
-            server = ldap3.Server(uri, get_info=ldap3.DSA)
+            server = ldap3.Server(url, get_info=ldap3.DSA)
             if bind:
                 self.conn = ldap3.Connection(server, user=bind,
                                              password=password)
--- a/vdirsyncer/storage/ldap.py
+++ b/vdirsyncer/storage/ldap.py
@@ -18,7 +18,6 @@ class LDAPStorage(Storage):
     :param filter: filter
     '''
     storage_name = 'ldap'
-    read_only = True
     fileext = '.vcf'
     item_mimetype = 'text/vcard'
 
@@ -107,3 +106,11 @@ class LDAPStorage(Storage):
         item = Item(vcard.serialize())
 
         return item, etag
+
+    def upload(self, item):
+        vcard = vobject.readOne(item.raw)
+        self.conn.strategy.add_entry('cn={},ou=test,o=lab'.format(vcard.fn), vcard)
+
+    def update(self, href, item, etag):
+        vcard = vobject.readOne(item.raw)
+        self.conn.strategy.add_entry(href, vcard)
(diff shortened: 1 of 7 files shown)


untitaker:
Cool! This looks reasonable, but:

* for VCard conversion please use https://github.com/eventable/vobject instead of your own thing, which actually has some bugs with values containing backslashes or newlines. This will also make it possible to parse VCards for uploading.
* Please specifiy `ldap` and `vobject` as dependencies here: https://github.com/pimutils/vdirsyncer/blob/master/setup.py#L69 For example, if I run `pip install vdirsyncer[remotestorage]`, `requests-oauthlib` will be installed. `ldap` and `vobject` should be installed when I run `pip install vdirsyncer[ldap]`.

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,84 @@
...
+
+from .base import Item, Storage
+from .. import exceptions
+
+ldap_logger = logging.getLogger(__name__)
+
+
+class LDAPStorage(Storage):
+
+    __doc__ = '''
```
untitaker: When implementing upload and update, we have to make it very clear which properties are even parsed and which ones are discarded.

untitaker:
And yes, please add tests for this, this is very meaningful! Do you have an idea how to do this? I.e., is there a simple ldap server we could use?

vdirsyncer/storage/ldap.py
untitaker: Also the `__doc__` is only necessary if we need to include variables from other strings.

paulfariello:
Done.

On Tue, Sep 13, 2016 at 07:55:19AM -0700, Markus Unterwaditzer wrote:
> Cool! This looks reasonable, but:
> 
> * for VCard conversion please use https://github.com/eventable/vobject instead of your own thing, which actually has some bugs with values containing backslashes or newlines. This will also make it possible to parse VCards for uploading.
> * Please specifiy `ldap` and `vobject` as dependencies here: https://github.com/pimutils/vdirsyncer/blob/master/setup.py#L69 For example, if I run `pip install vdirsyncer[remotestorage]`, `requests-oauthlib` will be installed. `ldap` and `vobject` should be installed when I run `pip install vdirsyncer[ldap]`.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-246709122

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

vdirsyncer/storage/ldap.py
paulfariello: Thanks for the quick reply.

I'm now using vobject which is clearly cleaner. Thanks for the hint.
I've also added support for title and fax.

I changed the default filter to one that make sense since `vobject`
clearly states that `F` and `FN` are mandatory fields. On some dirty
ldap directory `sn` or `givenName` could be missing and I'm not sure how to
handle it cleanly without filtering it. Maybe just provide the available
`family` or `givenname` for `N`.

About tests, I'm not sure how we could test it against a real ldap
server. I don't know any open ldap server but I can check.

Note that the ldap attributes I'm using are from Active Directory. Any
other ldap directory could also use it but this is not guaranteed. The
most problematic attribute is `whenChanged`. I've never seen it on other
classic ldap directory and it's used for etag.

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

paulfariello:
Thanks for the quick reply.

I'm now using vobject which is clearly cleaner. Thanks for the hint.
I've also added support for title and fax.

I changed the default filter to one that make sense since `vobject`
clearly states that `F` and `FN` are mandatory fields. On some dirty
ldap directory `sn` or `givenName` could be missing and I'm not sure how
to
handle it cleanly without filtering it. Maybe just provide the available
`family` or `givenname` for `N`.

About tests, I'm not sure how we could test it against a real ldap
server. I don't know any open ldap server but I can check.

Note that the ldap attributes I'm using are from Active Directory. Any
other ldap directory could also use it but this is not guaranteed. The
most problematic attribute is `whenChanged`. I've never seen it on other
classic ldap directory and it's used for etag.

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

untitaker:
It appears that this would be appropriate for testing: https://pypi.python.org/pypi/mockldap

See `tests/storage/test_filesystem.py` for how to write tests for the LDAP backend. In particular if you subclass `StorageTests` your storage will already be thoroughly tested, but `get_storage_args` has to be implemented.

untitaker:
> Note that the ldap attributes I'm using are from Active Directory. Any other ldap directory could also use it but this is not guaranteed. The most problematic attribute is `whenChanged`. I've never seen it on other classic ldap directory and it's used for etag.

Yeah I think we should make as much as possible configurable there, and also provide examples for other major LDAP providers.

untitaker:
Regarding `whenChanged`, if it's not there, you'll probably have to download the item and use `item.hash` or something similar.

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,97 @@
...
+    item_mimetype = 'text/vcard'
+
+    def __init__(self, uri=None, search_base=None, bind=None, password=None,
+                 filter='(&(objectCategory=person)(objectClass=user)'
+                        '(sn=*)(givenName=*))',
+                 **kwargs):
+        super(LDAPStorage, self).__init__(**kwargs)
+        self.search_base = search_base
+        self.filter = filter
+        self.server = ldap3.Server(uri)
```
untitaker: I think you could just use `ldap.initialize` here, no?

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,97 @@
...
+    :param bind: bind dn
+    :param password: bind password
+    :param filter: filter
+    '''
+    storage_name = 'ldap'
+    read_only = True
+    fileext = '.vcf'
+    item_mimetype = 'text/vcard'
+
+    def __init__(self, uri=None, search_base=None, bind=None, password=None,
```
untitaker: Why are all of these None by default? The ones that can't be optional shouldn't be.

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,97 @@
...
+                 filter='(&(objectCategory=person)(objectClass=user)'
+                        '(sn=*)(givenName=*))',
+                 **kwargs):
+        super(LDAPStorage, self).__init__(**kwargs)
+        self.search_base = search_base
+        self.filter = filter
+        self.server = ldap3.Server(uri)
+        if bind:
+            self.conn = ldap3.Connection(self.server, user=bind,
```
untitaker: What is the appropriate term, `user` or `bind`?

untitaker:
Am I correct to assume that this storage type does not support multiple addressbooks?

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,97 @@
...
+
+from .base import Item, Storage
+from .. import exceptions
+
+ldap_logger = logging.getLogger(__name__)
+
+
+class LDAPStorage(Storage):
+    '''
+    :param uri: LDAP URI
```
untitaker: Why URI instead of URL?

vdirsyncer/storage/ldap.py
untitaker: Ah, I see, you're using https://github.com/cannatag/ldap3, but I was reading the docs for https://github.com/pyldap/pyldap

Could you switch to the latter? This would make testing easier as `mockldap` appears to support `pyldap` only.

untitaker:
I added a LDAP testcase that should work if you use `pyldap` instead of `ldap3`.

paulfariello:
Thanks for the review. I'll fix today.

But first I have two questions:

- how am I supposed to run the test suites ? I'm running make -e test but got a bunch of errors
- how did you managed to push a commit on my fork ?

untitaker:
You need to run "make install-test" as well. In the documentation there is a page "Contributing" that goes into details.

Regarding how my commit landed on your fork, see https://github.com/blog/2247-improving-collaboration-with-forks

On 14 September 2016 09:23:24 CEST, Paul Fariello <notifications@github.com> wrote:
>Thanks for the review. I'll fix today.
>
>But first I have two questions:
>
>- how am I supposed to run the test suites ? I'm running make -e test
>but got a bunch of errors
>- how did you managed to push a commit on my fork ?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-246928119

-- 

paulfariello:
I've been trying to switch to pyldap but it doesn't natively support generator for ldap result. It makes it more complicated to cleanly integrate it.
Meanwhile I just discovered that ldap3 has a mocking component. I'll try to use it instead.

vdirsyncer/storage/ldap.py
paulfariello: I added a default value for uri to 'ldap://localhost'.
And try to get default value for base dn from server info.

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

vdirsyncer/storage/ldap.py
paulfariello: "Bind dn" is the appropriate term. See ldapsearch(1) for example.
You could bind to a ldap directory with a dn that is not a user.

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

paulfariello:
Yes

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

vdirsyncer/storage/ldap.py
paulfariello: URI is the right term for this since it can be an URL, a simple ip or a
hostname.

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

paulfariello:
Tests with ldap3 mocking are still WIP.

vdirsyncer/storage/ldap.py
untitaker: An IP or hostname is not a URI. URIs always have a scheme. https://tools.ietf.org/html/rfc3986#section-1.1.1

untitaker:
Ah, you need to add the new dependencies to `test-requirements.txt` as well.

paulfariello:
Ok you are right, I'll update the code
-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

paulfariello:
Should I add ldap3 in test-requirements.txt even if it's in setup.py ? 
-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

untitaker:
Yes because it's a optional requirement and otherwise not installed. This
situation needs a small refactor but not for now...

paulfariello:
Is it possible to use `StorageTests` with read only storage ?

untitaker:
No, unfortunately not. All it does is uploading items and then checking whether those can be downloaded again.


On 22 September 2016 09:00:46 CEST, Paul Fariello <notifications@github.com> wrote:
>Is it possible to use `StorageTests` with read only storage ?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-248827595

-- 

untitaker:
Are you interested in implementing `upload` and `update`? That would really make testing a lot easier (and I do think that is necessary).

paulfariello:
Main problem is that I'm not sure where to insert user on ldap directory and what fields are required.

But I'll try to add some basic update and upload. Anyway most ldap directory are read only for normal users.

Le 23 septembre 2016 10:58:51 GMT+02:00, Markus Unterwaditzer <notifications@github.com> a écrit :
>Are you interested in implementing `upload` and `update`? That would
>really make testing a lot easier (and I do think that is necessary).
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-249139442

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

untitaker:
I see. In that case it might make sense to write upload and update methods that just upload the data where the other storage methods can find it again, just for testing, but actually force the user to use read-only mode (I'd write this part)

On 23 September 2016 18:13:45 CEST, Paul Fariello <notifications@github.com> wrote:
>Main problem is that I'm not sure where to insert user on ldap
>directory and what fields are required.
>
>But I'll try to add some basic update and upload. Anyway most ldap
>directory are read only for normal users.
>
>Le 23 septembre 2016 10:58:51 GMT+02:00, Markus Unterwaditzer
><notifications@github.com> a écrit :
>>Are you interested in implementing `upload` and `update`? That would
>>really make testing a lot easier (and I do think that is necessary).
>>
>>-- 
>>You are receiving this because you authored the thread.
>>Reply to this email directly or view it on GitHub:
>>https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-249139442
>
>-- 
>Paul Fariello
>
>PGP: 0x672CDD2031AAF49B
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-249235546

-- 

untitaker:
@paulfariello BTW let me know if I can help with anything or if anything is unclear.

untitaker:
@paulfariello Any update on this? Adding tests is really important! Are you using this already for yourself?

paulfariello:
Sorry I've been quite busy lately. I'll try to add tests by the end of next week.

Le 8 octobre 2016 14:24:04 GMT+02:00, Markus Unterwaditzer <notifications@github.com> a écrit :
>@paulfariello Any update on this? Adding tests is really important! Are
>you using this already for yourself?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/pimutils/vdirsyncer/pull/500#issuecomment-252421803

-- 
Paul Fariello

PGP: 0x672CDD2031AAF49B

paulfariello:
I tried to add upload and update method. But I'm facing an issue with VCALENDAR. What should I do with it ? They doesn't really belong to an LDAP directory.

untitaker:
Fantastic, please push what you have, I'll do the changes such that VCALENDARs are not used in the tests.

paulfariello:
My last push is just a first attempt on update and upload implementation. It push to mocked server.
I also have an issue with start tls on mocked server.

untitaker:
Are `ldap://`-urls SSL by default?

untitaker:
I believe I've fixed the issue with ssl, let's just expect that the connection is fully set up if it's passed into the storage. I think ldap doesn't deal with SSL very well when mocked, but in our case SSL is not of any use anyway.

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+            vo = vcard.add('title')
+            vo.value = str(entry.title)
+
+        item = Item(vcard.serialize())
+
+        return item, etag
+
+    def upload(self, item):
+        vcard = vobject.readOne(item.raw)
+        self.conn.strategy.add_entry('cn={},ou=test,o=lab'.format(vcard.fn),
```
untitaker: Upload needs to return the assigned href and etag. It appears that your href is `dn`, but the etag has to be fetched like in `list` and `get`.

Also `add_entry` returns a boolean that indicates whether the entry has been added, you should check for that.

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+                         attributes=["whenChanged", "cn", "sn", "givenName",
+                                     "displayName", "telephoneNumber",
+                                     "mobile", "facsimileTelephoneNumber",
+                                     "mail", "title"])
+
+        if not self.conn.entries[0]:
+            raise exceptions.NotFoundError(href)
+
+        entry = self.conn.entries[0]
+        etag = str(entry.whenChanged)
```
untitaker: Here you have no fallback from `whenChanged` to `item.hash`. I suggest a separate helper function (outside of the storage class) for this.

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+
+        vcard = vobject.vCard()
+        vo = vcard.add('fn')
+        vo.value = str(entry.cn)
+        vo = vcard.add('n')
+        vo.value = vobject.vcard.Name(family=str(entry.sn),
+                                      given=str(entry.givenName))
+        if getattr(entry, 'telephoneNumber', None):
```
untitaker: This conversion from entry to item could be its own function (outside of storage class)

vdirsyncer/storage/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+        return item, etag
+
+    def upload(self, item):
+        vcard = vobject.readOne(item.raw)
+        self.conn.strategy.add_entry('cn={},ou=test,o=lab'.format(vcard.fn),
+                                     vcard)
+
+    def update(self, href, item, etag):
+        vcard = vobject.readOne(item.raw)
+        self.conn.strategy.add_entry(href, vcard)
```
untitaker: `add_entry` will do nothing if the item already exists, but if I call `update`, the item always exists. `update` is also expected to fail if the passed `etag` does not match the `etag` of the item on the server.

I think `modify` or how it's called is more appropriate than `add_entry`, but I'm not sure since I couldn't find much docs :(

untitaker:
Also please [contact](https://vdirsyncer.pimutils.org/en/stable/contact.html) me if there are any questions at all, PM, direct email, whatever too.

untitaker:
@paulfariello Ping!

untitaker:
Closing this due to inactivity, but for reference, this is a good start for a successor PR.

PR #507, opened 2016-09-19
Strip trailing slash from .well-known-URIs to fix autodiscovery

citrin:
According to RFC6764 URLs should be "/.well-known/caldav" and
"/.well-known/caldav".

Tested with Baïkal, but redirect for .well-known-URIs defined in nginx config so it is not Baïkal-specific.

May be `https://www.googleapis.com/.well-known/carddav/` should also be changed to `https://www.googleapis.com/.well-known/carddav`. See: https://developers.google.com/google-apps/carddav/ But I don't use vdirsyncer with google and can't test this change.

### DIFF
--- a/vdirsyncer/storage/dav.py
+++ b/vdirsyncer/storage/dav.py
@@ -281,7 +281,7 @@ class CalDiscover(Discover):
     </d:propfind>
     """
     _homeset_tag = '{%s}calendar-home-set' % _namespace
-    _well_known_uri = '/.well-known/caldav/'
+    _well_known_uri = '/.well-known/caldav'
 
 
 class CardDiscover(Discover):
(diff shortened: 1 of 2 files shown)


untitaker:
I'd be fine with the fix for Google as well. Is there a practical issue you're fixing with this change?

citrin:
> I'd be fine with the fix for Google as well

I think trailing slash should be removed from well-known URL for google too, but it is not easy to test.

> Is there a practical issue you're fixing with this change?

I use Baïkal with nginx and autodiscovery doesn't work for me without this change. Well-known URLs are defined in nginx config and I can change nginx config, but it is better to comply with RFC, especially when it is easy.

untitaker:
Just go for the change of Google storage, it'll be fine!

untitaker:
Groovy, thanks!

PR #512, opened 2016-09-29
fix yaml syntax

stevepeak:
This will fix the status for sure. The comment may be fixed as well since the yaml is now valid

> CC https://github.com/codecov/support/issues/162#issuecomment-250472579

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,4 @@
 comment: false
-status:
+coverage:
+  status:
     patch: false


untitaker:
Which filename is correct `.codecov.yml` or `codecov.yml`?

untitaker:
Also thanks for the PR

stevepeak:
Both path names will work. 👍

untitaker:
Hmm, under https://codecov.io/gh/pimutils/vdirsyncer/settings/yaml I still see the default config @stevepeak

PR #515, opened 2016-10-03
Partial sync

### LINKED ISSUE #468: partial sync
Is there a way to do unidirectional partial sync with vdirsyncer?

That is, considering to calendars A and B: 

* all events of A are put in B
* if an event is moved in A it is moved in B
* if an event is deleted in A it is deleted in B
* if an event coming from A is moved or deleted in B, an error is displayed
* (B may contain events that are not in A)

untitaker:
Fix #468

- [x] cli
- [x] docs
- [x] changelog


untitaker:
I decided not to make partial sync an error by default because that would be a breaking API change.

PR #529, opened 2016-12-15
Fix log debug exceptions

Nicop06:
This will avoid the ugly stack when activating debug log.

### DIFF
--- a/vdirsyncer/utils/http.py
+++ b/vdirsyncer/utils/http.py
@@ -49,8 +49,8 @@ def request(method, url, session=None, latin1_fallback=True,
     func = session.request
 
     logger.debug(u'{} {}'.format(method, url))
-    logger.debug(kwargs.get('headers', {}))
-    logger.debug(kwargs.get('data', None))
+    logger.debug(str(kwargs.get('headers', {})))
+    logger.debug(kwargs.get('data', None).decode('utf8'))
     logger.debug('Sending request...')
 
     assert isinstance(kwargs.get('data', b''), bytes)
@@ -65,9 +65,9 @@ def request(method, url, session=None, latin1_fallback=True,
         logger.debug('Removing latin1 fallback')
         r.encoding = None
 
-    logger.debug(r.status_code)
-    logger.debug(r.headers)
-    logger.debug(r.content)
+    logger.debug(str(r.status_code))
+    logger.debug(str(r.headers))
+    logger.debug(r.content.decode('utf8'))
 
     if r.status_code == 412:
         raise exceptions.PreconditionFailed(r.reason)


Nicop06:
I know I didn't add myself to the change log and author files as I consider this a minor fix. If you still want me to do so please let me know.

untitaker:
Which stack do you see? For me it shows the repr for non-str types (i.e. str() is called from within logging)

Nicop06:
This problem is actually in `click_log`. It is fixed in master but not 0.1.5 (the latest release). I suppose you use the development version, that's why you don't see the stack.

Here is the fix: <https://github.com/click-contrib/click-log/blame/master/click_log/core.py#L34>. Notice the `splitlines` called directly on the object given to the logger.

You can close this pull request as this is not vdirsyncer related. I should have looked a bit further. Although, using the encode function on binary data gives much better as `\n` are interpreted as `EOL`, so it could be interested to patch it. I can open another PR or update this one if you want.

untitaker:
click-log 0.1.5 contains that fix, the release was specifically made because of that fix.

Can you confirm that it works with latest click-log?

untitaker:
Independently of that, I never saw a stacktrace there, regardless of click-log's version.

Nicop06:
Sorry, I was wrong. Actually click-log 0.1.5 introduces the regression with this commit:

https://github.com/click-contrib/click-log/commit/5a1fdc4f599a83965d242717a9a15a2d99c63579

And here is the stack:

```
Traceback (most recent call last):
  File "/home/nporcel/.virtualenvs/vdirsyncer/lib/python3.4/site-packages/click_log/core.py", line 42, in emit
    msg = self.format(record)
  File "/usr/lib/python3.4/logging/__init__.py", line 828, in format
    return fmt.format(record)
  File "/home/nporcel/.virtualenvs/vdirsyncer/lib/python3.4/site-packages/click_log/core.py", line 34, in format
    for x in record.msg.splitlines())
  File "/home/nporcel/.virtualenvs/vdirsyncer/lib/python3.4/site-packages/click_log/core.py", line 34, in <genexpr>
    for x in record.msg.splitlines())
```

This is the small stack, as it is raised again in the emit function. But basically the full stack contains the line I modified in `vdirsyncer`.

untitaker:
I released 0.1.7 of click-log, does this fix the issue?

Nicop06:
This is better, but I have this error now:

```
TypeError: Can't convert 'bytes' object to str implicitly
```

Basically, this is this expression that fails: `prefix + x`, as `prefix` is string and `x` is of type `bytes`. 

I opened a pull request: https://github.com/click-contrib/click-log/pull/3

untitaker:
Could you try click-log from the master branch? It should fix this bug once and for all and make https://github.com/click-contrib/click-log/pull/3 obsolete.

Nicop06:
I cannot test it before Monday. But looking at the code and the tests it seems perfectly fine so I closed the both this PR and https://github.com/click-contrib/click-log/pull/3. Good thinking on the python2 test.

PR #558, opened 2017-02-16
Add trove classifiers

### LINKED ISSUE #557: Does this package support Python 3?
On 19 February 2017, Python 3 will be 3,000 days old!  We are interested to see if we can get **at least 50%** of the [Top 5,000 PyPI packages](http://python3wos.mybluemix.net/5000) to compatible with Python 3 by that date.  We are *really* close and given that this package is in the PyPI Top 5,000, we seek your assistance in pushing us over that threshold.

  So, if this package *already supports Python 3* then please consider adding an appropriate [trove classifier](https://pypi.python.org/pypi?%3Aaction=list_classifiers) "*Programming Language :: Python :: 3*" to this package's PyPI page so that tools can automaticly determine its Python 3 compatibility.

hobarrera:
Fixes #557

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,7 @@ setup(
         'console_scripts': ['vdirsyncer = vdirsyncer.cli:main']
     },
     classifiers=[
-        'Development Status :: 5 - Production/Stable',
+        'Development Status :: 4 - Beta',
         'Environment :: Console',
         'License :: OSI Approved :: MIT License',
         'Operating System :: POSIX',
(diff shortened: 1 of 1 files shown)


hobarrera:
Not 100% sure if we want `Topic :: Office/Business :: Groupware`. Also, not sure if `Stable` or `Mature`.

untitaker:
This is definetly not a groupware. I don't want to state that this is either stable or mature.

hobarrera:
Sorry, I confused groupware with something else (PIM). 4/Beta then?

untitaker:
Yes I think so.

PR #559, opened 2017-02-17
Adding tip about Microsoft's Remove-DuplicateAppointments.ps1

somatronicon:
Adding some documentation around what users can do if they encounter duplicate events using DavMail.

### DIFF
--- a/docs/supported.rst
+++ b/docs/supported.rst
@@ -218,6 +218,26 @@ Exchange server you might get confronted with weird errors of all sorts
 
 .. _DavMail: http://davmail.sourceforge.net/
 
+**If you run into duplicate events**::
+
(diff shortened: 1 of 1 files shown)


untitaker:
I'm curious, does the command suggested by vdirsyncer work in such cases as well?

somatronicon:
Hi @untitaker, thanks for the response. I did try to perform the recovery a few times prior to using the PowerShell script, and I got exceptions each time I attempted the recovery. I tried with the unsafe flag once or twice as well. As the instructions suggested in the output, I assumed it was more a problem with the server software than a problem with vdirsyncer, so I started looking for ways I could clean things up on the server; that's when I stumbled onto the PowerShell script.

untitaker:
Ok Thanks!

PR #562, opened 2017-02-21
Enable PyPy3

### LINKED ISSUE #429: Test against pypy Python 3 in CI

untitaker:
Fix #429

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -166,6 +166,30 @@
         "python": "3.6"
       },
       {
+        "env": "BUILD=test DAV_SERVER=skip REQUIREMENTS=devel ",
+        "python": "pypy3"
+      },
+      {
+        "env": "BUILD=test DAV_SERVER=skip REQUIREMENTS=release ",
+        "python": "pypy3"
+      },
+      {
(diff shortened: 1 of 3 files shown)


hobarrera:
Since I'd like to imitate this for todoman, I'd like to understand this further; you're ignoring the travis-provided pypy, and installing it manually via `pyenv`, correct?

I take it the travis-provided one didn't work?

untitaker:
The Travis-provided one was too old, I think it targeted Python 3.2. Theoretically we could make vdirsyncer 3.2-compatible now that we don't support Python 2 anymore (we relied on usage of `u""`-strings, which are only available in Py2 and Py3.3+), but I don't have the incentive to do that.

untitaker:
Also yes we're installing a newer pypy3 into `$HOME` (not system-wide), but without removing the old one from travis.

hobarrera:
Hmm. Thanks. todoman's travis setup is a lot simpler and I'd rather wait for travis to update and not make it more complex.

PR #563, opened 2017-02-22
Shell-expand user home directory in confres command

madduck:
Let me specify the command to use for conflict resolution with `~` in the path, which gets expanded to my homedir.

Signed-off-by: martin f. krafft <madduck@madduck.net>

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -11,5 +11,6 @@ In alphabetical order:
 - Malte Kiefer
 - Marek Marczykowski-Górecki
 - Markus Unterwaditzer
+- Martin Krafft
 - Michael Adler
 - Thomas Weißschuh
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,9 @@ Version 0.15
 - Deprecated syntax for configuration values is now completely rejected. All
   values now have to be valid JSON.
 - A few UX improvements for Google storages, see :gh:`549` and :gh:`552`.
+- Conflict resolution commands now get path-expanded, i.e. "~/bin/myresolver"
+  is now acceptable, whereas before one had to specify the absolute path
+  including the home directory, see :gh:`563`.
 
 Version 0.14.1
 ==============
--- a/vdirsyncer/cli/config.py
+++ b/vdirsyncer/cli/config.py
@@ -246,6 +246,8 @@ class PairConfig(object):
             def resolve(a, b):
                 a_name = self.config_a['instance_name']
                 b_name = self.config_b['instance_name']
+                conflict_resolution[1] = \
+                        os.path.expanduser(conflict_resolution[1])
                 command = conflict_resolution[1:]
 
                 def inner():


untitaker:
Are you sure this works? It appears that you're passing a list of strings to expanduser.

madduck:
@untitaker again, hawkeyes… sorry that was sloppy. I tested this, then lost my changes to git, and then just did it again without re-testing.

It's now updated and should be a better solution anyway.

untitaker:
LGTM, please

* add a changelog entry
* add a testcase somewhere around https://github.com/pimutils/vdirsyncer/blob/master/tests/system/cli/test_fetchparams.py#L41
* add yourself to AUTHORS

madduck:
I've added the changelog entry, but I am having troubles with the test case. Seems like I should just derive from `test_conflict_resolution` (i.e. follow a similar pattern), but since that actually runs a sync designed to call the resolution script, I'd have to write something into the home directory of the running user, and I am not comfortable doing that.

I'm not familiar with the test framework you're using. Does it allow mock-ups, i.e. in this case, could I mock up the `ResolveConflict` instance that gets used and make the test inspect the `run` call's `conflict_resolution` argument to check whether the path was properly expanded?

untitaker:
Indeed I forgot about that, and it seems I even linked the wrong file earlier. If you move your `expanduser` call into `_resolve_conflict_via_command` you should be able to write a simple "unit"test that doesn't require any mocking. Please create a new file called `tests/unit/cli/test_config.py` for this test.

madduck:
Really, @untitaker? If I move the `expanduser` call into `_resolve_conflict_via_command` as you suggest, wouldn't I still have to mock up `subprocess.check_call`?

Or are you suggesting that the test creates an impossible path, such as `~/.` which will cause `subprocess.check_call` to throw a `PermissionError`, so if I catch that within the function, add the filename to the error object and re-raise, then I could catch that from the test, check the filename in the exception object and let the test succeed if the expansion happened?

madduck:
Another option I guess would be to create a script in a temporary file and then use `os.path.relpath` and `os.path.expenaduser` in the test to generate a path such as `~/../../tmp/vdirsyncer-conflict.foo/testscript.sh`, which ought to work. I just question the utility of this, somewhat.

untitaker:
Sorry, yes you'd need to monkeypatch anyway. You can do that with the pytest monkeypatch fixture.

Still better than triggering an entire sync

On 23 February 2017 01:55:17 CET, "martin f. krafft" <notifications@github.com> wrote:
>Really, @untitaker? If I move the `expanduser` call into
>`_resolve_conflict_via_command` as you suggest, wouldn't I still have
>to mock up `subprocess.check_call`?
>
>Or are you suggesting that the test creates an impossible path, such as
>`~/.` which will cause `subprocess.check_call` to throw a
>`PermissionError`, so if I catch that within the function, add the
>filename to the error object and re-raise, then I could catch that from
>the test, check the filename in the exception object and let the test
>succeed if the expansion happened?
>
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/pimutils/vdirsyncer/pull/563#issuecomment-281855293

-- 

madduck:
This is a bit over my head as I don't know what you want me to actually test. `PairConfig` seems to not have a public API, at least not one that touches `self.conflict_resolution`.

Do you want me to test the function `_resolve_conflict_via_command` directly, after monkey-patching the `execvp` call?

I'll have to read up on how to do this and can't continue now anyway. Sorry.

untitaker:
>Do you want me to test the function _resolve_conflict_via_command directly, after monkey-patching the execvp call?

I don't think you'd have to patch on such a low level. I was thinking of patching `subprocess.Popen` (assuming `check_call` uses that, otherwise patch `check_call` itself) using [pytest's monkeypatching support](http://doc.pytest.org/en/latest/monkeypatch.html)

So you'd need to add a function parameter `monkeypatch` to your test, which is then recognized by pytest as some sort of import, and then use the following line once:

    monkeypatch.setattr("subprocess.Popen", my_custom_popen)

Cleanup will happen automatically after the test, so calling `monkeypatch.undo()` or similar is not necessary.

You can just raise an exception from within there to avoid returning a useful value, since this is just for validating whether it has been correctly called. To catch that exception you can use [`pytest.raises`](http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) (a normal function you'd have to normally import).

untitaker:
Hey @madduck are there any updates on this? Thanks!

madduck:
No, @untitaker sorry. I've been swamped with family and now work. I wasn't aware that this tiny patch would turn into such a deal… :/

untitaker:
to be clear, are you still interested in working on this in the near future? Otherwise I suppose I'll fix it.

madduck:
If it's a 5 minute job for you, please do it. I never shy away from little tasks like this, but I just don't see myself invest the half hour it'd take me anytime soon. Sorry!

PR #572, opened 2017-03-01
Add fastmail testserver

### LINKED ISSUE #571: Test Cyrus in Travis
[Cyrus](http://cyrusimap.org/) is a IMAP, CalDAV and CardDAV server, used by FastMail. We should test it in Travis.

untitaker:
Fix #571

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -13,3 +13,6 @@ In alphabetical order:
 - Markus Unterwaditzer
 - Michael Adler
 - Thomas Weißschuh
+
+Additionally `FastMail sponsored a paid account for testing
+<https://github.com/pimutils/vdirsyncer/issues/571>`_. Thanks!
--- a/tests/storage/__init__.py
+++ b/tests/storage/__init__.py
@@ -250,8 +250,8 @@ class StorageTests(object):
                           get_storage_args, get_item):
         if getattr(self, 'dav_server', '') == 'radicale':
             pytest.skip('Radicale is fundamentally broken.')
-        if getattr(self, 'dav_server', '') == 'icloud':
-            pytest.skip('iCloud rejects uploads.')
+        if getattr(self, 'dav_server', '') in ('icloud', 'fastmail'):
+            pytest.skip('iCloud and FastMail reject this name.')
 
         monkeypatch.setattr('vdirsyncer.utils.generate_href', lambda x: x)
 
--- a/tests/storage/__init__.py
+++ b/tests/storage/__init__.py
@@ -310,7 +310,7 @@ class StorageTests(object):
         if item_type != 'VEVENT':
             pytest.skip('This storage instance doesn\'t support iCalendar.')
 
-        uid = u'abc123'
+        uid = str(uuid.uuid4())
         item = Item(textwrap.dedent(u'''
         BEGIN:VCALENDAR
         VERSION:2.0
--- a/tests/storage/__init__.py
+++ b/tests/storage/__init__.py
@@ -203,8 +203,9 @@ class StorageTests(object):
         if getattr(self, 'dav_server', '') == 'radicale':
             pytest.skip('MKCOL is broken under Radicale 1.x')
 
-        if getattr(self, 'dav_server', '') == 'icloud':
-            pytest.skip('iCloud requires a minimum-length for collection name')
+        if getattr(self, 'dav_server', '') in \
+           ('icloud', 'fastmail', 'davical'):
+            pytest.skip('Manual cleanup would be necessary.')
 
         args = get_storage_args(collection=None)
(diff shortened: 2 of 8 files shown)


tests/storage/servers/fastmail/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+import pytest
+
+args = {
+    'username': os.environ['FASTMAIL_USERNAME'],
+    'password': os.environ['FASTMAIL_PASSWORD']
+}
+
+carddav_args = dict(args)
+carddav_args['url'] = (
+    'https://carddav.messagingengine.com/dav/addressbooks/user/{}/'
```
untitaker: This URL is very old, we should probably try to simplify it.

AUTHORS.rst
```diff
@@ -13,3 +13,6 @@ In alphabetical order:
 - Markus Unterwaditzer
 - Michael Adler
 - Thomas Weißschuh
+
+Additionally `FastMail sponsored a paid account for testing
```
hobarrera: A _comped_ account, actually.

untitaker: I don't really know what that means.

hobarrera: >  Something of value, such as a ticket for admission or a book, given free of charge
[source](http://www.thefreedictionary.com/comped).

I didn't either, but that's the wording used for the offer. :)

untitaker: I'll leave it at the current wording assuming there's no difference in meaning.

untitaker:
Had to kick Travis over due to partial outage on their side

AUTHORS.rst
hobarrera: Yeah, sure, it's better to leave it with what you're comfortable with.

untitaker:
ok, travis is just completely broken. I'll wait until their status turns green again

untitaker:
Deleting collections on teardown is not working properly.

tests/storage/__init__.py
```diff
@@ -203,8 +203,9 @@ def test_create_collection(self, requires_collections, get_storage_args,
         if getattr(self, 'dav_server', '') == 'radicale':
             pytest.skip('MKCOL is broken under Radicale 1.x')
 
-        if getattr(self, 'dav_server', '') == 'icloud':
-            pytest.skip('iCloud requires a minimum-length for collection name')
+        if getattr(self, 'dav_server', '') in \
+           ('icloud', 'fastmail', 'davical'):
+            pytest.skip('Manual cleanup would be necessary.')
```
hobarrera: I managed to create and delete a calendar on Fastmail from my iPhone (CalDAV). Here's the request from `mitmproxy`:

```
DELETE https://caldav.fastmail.com/dav/calendars/user/hobarrera%40fastmail.com/9B0C98FD-6E99-4684-80FD-3198E43F865D/

Host:             caldav.fastmail.com
Authorization:    Basic REDACTED
Connection:       keep-alive
Accept:           */*
User-Agent:       iOS/10.2.1 (14D27) dataaccessd/1.0
Content-Length:   0
Accept-Language:  en-gb
Accept-Encoding:  gzip, deflate
```

tests/storage/__init__.py
```diff
@@ -249,8 +250,8 @@ def test_specialchars(self, monkeypatch, requires_collections,
                           get_storage_args, get_item):
         if getattr(self, 'dav_server', '') == 'radicale':
             pytest.skip('Radicale is fundamentally broken.')
-        if getattr(self, 'dav_server', '') == 'icloud':
-            pytest.skip('iCloud rejects uploads.')
+        if getattr(self, 'dav_server', '') in ('icloud', 'fastmail'):
+            pytest.skip('iCloud and FastMail reject this name.')
```
hobarrera: Which name is rejected? I might be able to get in touch and see if it's a bug on their side.

untitaker:
I'll investigate later, very busy right now. `DELETE` works if I do it manually, and I do it the same way in the tests. Something else is going wrong.

tests/storage/__init__.py
untitaker: That's not the issue, we would need to have a `delete` method across storages, although we don't have a need for this during normal usage. Also creating and deleting collections is tested with iCloud and FastMail all the time, (because that's the only way to properly set up and tear down all state) so there's really no need for this test to be run.

tests/storage/__init__.py
untitaker: It would be totally understandable if they intentionally reject this upload, it amounts to fuzzing what I do here.

untitaker: >Which name is rejected?

Creating a collection with the GUID `test @ foo ät bar`.

hobarrera:
🎉

PR #573, opened 2017-03-02
config.example: Status path needs to be quoted

tbabej:
Seems like this one is missing quotation (and I was unable to get vdirsyncer working without it).


untitaker:
Thanks!

PR #587, opened 2017-03-07
Fixed some typos.

geier:
I installed mwic, so you don't have to.

`unperformant` doesn't seem to be an often used english word either and
`addressbook` is probably `address book`.


untitaker:
Thanks!

PR #593, opened 2017-03-09
Use MKCALENDAR

### LINKED ISSUE #570: Investigate switch from MKCOL to MKCALENDAR for caldav.
When synchronizing calendars with iCloud using the `caldav` storage, created collections don't support VTODO entries. As outlined in https://github.com/pimutils/vdirsyncer/pull/567#issuecomment-283036234 this may be fixed by using MKCALENDAR instead of MKCOL to create calendars.

untitaker:
Fix #570

### DIFF
--- a/tests/storage/servers/icloud/__init__.py
+++ b/tests/storage/servers/icloud/__init__.py
@@ -7,11 +7,6 @@ class ServerMixin(object):
 
     @pytest.fixture
     def get_storage_args(self, item_type, slow_create_collection):
-        if item_type != 'VEVENT':
-            # For some reason the collections created by vdirsyncer are not
-            # usable as task lists.
-            pytest.skip('iCloud doesn\'t support anything else than VEVENT')
-
         def inner(collection='test'):
             args = {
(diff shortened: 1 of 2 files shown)


hobarrera:
I'd also include a test uploading a vcard to iCloud on this PR to confirm that this has the desired effect (I suspect it does).

untitaker:
For now I'm just trying to get the CI to pass

untitaker:
iCloud only lets us create a collection that is *either* a calendar *or* a tasklist. See https://travis-ci.org/pimutils/vdirsyncer/jobs/210239842#L782:

    debug: <html><head><title>Bad Request</title></head><body><h1>Bad Request</h1><p>Calendar SupportedCalendarComponentSet length exceeded, only one calendar component type supported</p></body></html>

Vdirsyncer could support an additional storage param that decides what to create, but I think it's just way easier if users created collections in the web interface.

hobarrera:
Wow, that's really bad UX (on iCloud's side), since the iOS app lets me create both VTODOs and VEVENTs on the same calendar. I wonder how they deal with that when syncing with iCloud. :/

PR #599, opened 2017-03-14
Use MKCALENDAR to create calendar collections

hobarrera:
This adds support for calendar creation for radicale.

TBH, I'm not sure if this will work, and travis is the fastest way to test it.

### DIFF
--- a/vdirsyncer/storage/dav.py
+++ b/vdirsyncer/storage/dav.py
@@ -302,6 +302,30 @@ class CalDiscover(Discover):
     _homeset_tag = '{%s}calendar-home-set' % _namespace
     _well_known_uri = '/.well-known/caldav'
 
+    def _create_collection_impl(self, url):
+        data = '''<?xml version="1.0" encoding="utf-8" ?>
+        <C:mkcalendar xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav">
+            <D:set>
+                <D:prop>
+                <D:displayname>{}</D:displayname>
+                <C:supported-calendar-component-set>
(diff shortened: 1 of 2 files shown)


vdirsyncer/storage/dav.py
```diff
@@ -302,6 +302,30 @@ class CalDiscover(Discover):
...
+                <D:displayname>{}</D:displayname>
+                <C:supported-calendar-component-set>
+                    <C:comp name="VEVENT"/>
+                    <C:comp name="VTODO"/>
+                </C:supported-calendar-component-set>
+                </D:prop>
+            </D:set>
+        </C:mkcalendar>'''.format(
+            self._get_collection_from_url(url)
+        ).encode()
```
untitaker: Please use `utf-8` here.

hobarrera: That's the default: https://docs.python.org/3/library/stdtypes.html#str.encode

hobarrera: Or did you mean that you prefer it explicitly stated?

hobarrera:
I've no idea why this failed so horribly. It's based on your own MKCALENDAR branch. :(

vdirsyncer/storage/dav.py
untitaker: I didn't know that, but I think I prefer it to be explicit since I switch between Py2 and Py3 all the time.

untitaker:
@hobarrera can we close this?

hobarrera:
Yeah, it's a bit beyond me right now :(

PR #601, opened 2017-03-14
Add xandikos to CI

### LINKED ISSUE #597: Add Xandikos to CI
#596 triggered my interest in Xandikos, so it might be worth adding it to CI.

[Its site](https://jelmer.uk/code/xandikos/) actually lists vdirsyncer as compatible! :+1:

I'd like to tackle this myself, but would actually like to see if a free-tier heroku "machine" can run Xandikos. We might be able to avoid the bootstrapping/installation overhead.

untitaker:
Fix #597

Bugs:
- [x] https://github.com/jelmer/xandikos/pull/18
- [x] ~~`calendar-home-set` is missing, so blocked on https://github.com/jelmer/xandikos/issues/17~~ Nevermind, I'm an idiot
- [x] REPORT doesn't work properly: https://github.com/jelmer/xandikos/issues/16
- [x] REPORT doesn't work properly v2: https://github.com/jelmer/xandikos/issues/30
- [x] Validity of items: https://github.com/jelmer/xandikos/issues/24
- [x] Support for If-None-Match: https://github.com/jelmer/xandikos/issues/23
- [x] Trailing slashes: jelmer/xandikos#42

Other chores:
- [x] Add to list of supported servers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,15 +34,7 @@
         "python": "3.3"
       },
       {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=release BUILD_PRS=true ",
-        "python": "3.3"
-      },
-      {
         "env": "BUILD=test DAV_SERVER=radicale REQUIREMENTS=minimal BUILD_PRS=true ",
-        "python": "3.3"
-      },
-      {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=minimal BUILD_PRS=true ",
         "python": "3.3"
       },
       {
@@ -58,15 +50,7 @@
         "python": "3.4"
       },
       {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=release BUILD_PRS=true ",
-        "python": "3.4"
-      },
-      {
         "env": "BUILD=test DAV_SERVER=radicale REQUIREMENTS=minimal BUILD_PRS=true ",
-        "python": "3.4"
-      },
-      {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=minimal BUILD_PRS=true ",
         "python": "3.4"
       },
       {
@@ -82,15 +66,7 @@
         "python": "3.5"
       },
       {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=release BUILD_PRS=true ",
-        "python": "3.5"
-      },
-      {
         "env": "BUILD=test DAV_SERVER=radicale REQUIREMENTS=minimal BUILD_PRS=true ",
-        "python": "3.5"
-      },
-      {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=minimal BUILD_PRS=true ",
         "python": "3.5"
       },
       {
@@ -102,15 +78,7 @@
         "python": "3.6"
       },
       {
-        "env": "BUILD=test DAV_SERVER=skip REQUIREMENTS=release BUILD_PRS=true ",
-        "python": "3.6"
-      },
-      {
         "env": "BUILD=test DAV_SERVER=radicale REQUIREMENTS=release BUILD_PRS=true ",
-        "python": "3.6"
-      },
-      {
-        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=release BUILD_PRS=true ",
         "python": "3.6"
       },
       {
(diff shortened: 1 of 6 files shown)


untitaker:
Xandikos 404s when trying to list all collections, also it does not properly differentiate between `addressbook-home-set` and `calendar-home-set`, which means that MKCOL on those URLs will not work properly since the server won't know which collection to create.

This is RFC-conformant but unusuable behavior. Same problem as Radicale.

hobarrera:
> Xandikos 404s when trying to list all collections

Oh, I thought I was doing something wrong when I approached this, so it wasn't my bad. :(

> also it does not properly differentiate between addressbook-home-set and calendar-home-set, which means that MKCOL on those URLs will not work properly since the server won't know which collection to create.

I'd like to understand this a bit more (and also, forgive my slight ignorance here), but, what don't we includ `DAV:resourcetype` in `MKCOL`? Shouldn't that suffice to identify the collection type?

untitaker:
IIRC MKCOL with request body was rejected by other servers.

jelmer:
Listing calendars with vdirsyncer works fine against xandikos for me; what request 404s for you?

MKCOL with a body indeed allows specifying what resource types the new collection should have. That said, Xandikos doesn't yet support setting the resource types of newly created collections. https://github.com/jelmer/xandikos/issues/16. It's correct that some servers reject a MKCOL with body; servers should reject a body they don't understand per RFC4918; RFC5689 defines an XML body for MKCOL.

I've filed https://github.com/jelmer/xandikos/issues/21 about automatically setting the type of collections based on under what path they were created, if there was no body specified for MKCOL.

untitaker:
I think there's a misunderstanding in how I should set up xandikos for testing, could you review this PR?

untitaker:
@jelmer Actually we send a MKCOL with all the properties that are necessary for xandikos to infer the collection type: https://github.com/pimutils/vdirsyncer/blob/master/vdirsyncer/storage/dav.py#L266

untitaker:
Turns out I wasn't setting up Xandikos correctly. We're now able to create collections properly, but the tests still don't pass. I will file issues on Xandikos for that.

jelmer:
With current vdirsyncer and current xandikos, all tests now pass for me.

.travis.yml
```diff
@@ -26,42 +26,82 @@
...
+        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=release BUILD_PRS=true ",
+        "python": "3.3"
+      },
+      {
         "env": "BUILD=test DAV_SERVER=radicale REQUIREMENTS=minimal BUILD_PRS=true ",
         "python": "3.3"
       },
       {
+        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=minimal BUILD_PRS=true ",
+        "python": "3.3"
```
hobarrera: Xandikos only support python 3.5 or above:

https://github.com/jelmer/xandikos#dependencies

untitaker: @jelmer Would you consider supporting Python 3.3+ (i.e. I submit PRs to xandikos if Python 3.3 breaks)

jelmer:
On Sun, Mar 19, 2017 at 06:18:18AM -0700, Markus Unterwaditzer wrote:
> untitaker commented on this pull request.
> 
> 
> 
> >          "env": "BUILD=test DAV_SERVER=radicale REQUIREMENTS=minimal BUILD_PRS=true ",
>          "python": "3.3"
>        },
>        {
> +        "env": "BUILD=test DAV_SERVER=xandikos REQUIREMENTS=minimal BUILD_PRS=true ",
> +        "python": "3.3"
> 
> @jelmer Would you consider supporting Python 3.3+ (i.e. I submit PRs to xandikos if Python 3.3 breaks)
For 3.4 that would probably be okay so long as it doesn't introduce
too much extra complexity.

3.3 is trickier, since e.g. Dulwich also doesn't support 3.3 at the
moment.

jelmer:
Dulwich & Xandikos now support 3.3. That was way easier than I was expecting, no changes necessary.

untitaker:
@jelmer the tests are failing for anything below 3.6. It appears there is a bug in the urllib module that makes xandikos output double slashes for `calendar-home-set`

untitaker:
One of your recent commits breaks the CI again :(

jelmer:
Argh, fixed now.

Another reason to extend the unit test coverage further...

hobarrera:
I wonder if, as a way to get around the temporarily lack of coverage, you could run vdirsyncer's test suite for Xandiko's CI. :P

hobarrera:
:tada:

PR #602, opened 2017-03-14
Add quotes to the suggested config

kuba-orlik:
Without the quotes the app doesn't start, throwing: 

```
critical: Error during reading config /home/kuba/.vdirsyncer/config: Section "general": Section "general", option "status_path": Expecting value: line 1 column 1 (char 0)
```


untitaker:
Thanks!

PR #608, opened 2017-03-22
Use py.test-3 if available.

jelmer:
py.test on some platforms uses python 2.


untitaker:
I'd rather not do this, but assume the user used pip to install pytest in a virtualenv. I've also seen pytest3 as a name given by distro packages.

hobarrera:
Likewise -1 for the same reasons.

jelmer:
Followed up on https://github.com/jelmer/xandikos/pull/45

Issue #546, opened 2017-02-06
Switch status format to sqlite


untitaker:
This should've been obvious from the beginning (given offlineimap's development), but I guess it's never too late: We should switch to sqlite for storing the status metadata in `<statusfolder>/<pair>/<collection>.items`.

* `.collections` files should still be JSON, only the format for `.items` changes. At least I see no reason to do otherwise.
* The codebase for loading and saving `.collections` and `.items` files is weirdly entangled. We need to fix that.
* `sync` should take subclass of a `Status` interface. Requirement to remove coupling to either in-memory dict or sqlite as status storage, because otherwise we can't write decent unit tests. 

Goal: Ability to modify status in-place (instead of having "load" and "save" steps) opens the door to optimizing memory-efficiency of sync. I played with this once and it's not easy. We'll see.

hobarrera:
Any reason for the collections file to still be json (rather than sql)? I feel like having a single caching format might simplify code with no real downside (though I may be generalizing a bit).

untitaker:
I don't particularly care to be honest. Whatever turns out the easiest.

untitaker:
Setting this to blocker since I pushed a half-assed implementation to master.

PR #609, opened 2017-03-25
Use sqlite for sync internally

untitaker:
Another preparation for #546

This loads the legacy JSON format into a in-memory sqlite db, does some
operations there and converts it back into JSON at the end. That's not the end
goal, but makes testing easier since we don't have to modify any tests. Later
we can then modify `sync` to finally accept a sqlite db object and do the
necessary modifications in `vdirsyncer.cli`.

Right now there are two correctness bugs left though.

### DIFF
--- a/vdirsyncer/sync.py
+++ b/vdirsyncer/sync.py
@@ -304,7 +304,7 @@ class _SqliteStatus(_StatusBase):
         if res is None:
             return None
 
-        if res['hash'] is None:
+        if res['hash'] is None:  # FIXME: Implement as constraint in db
             assert res['href'] is None
             assert res['etag'] is None
             return None
--- a/vdirsyncer/sync.py
+++ b/vdirsyncer/sync.py
@@ -358,10 +358,12 @@ class _SqliteStatus(_StatusBase):
         )
 
     def get_by_href_a(self, *a, **kw):
-        return self._get_by_href_impl(*a, **kw, side='a')
+        kw['side'] = 'a'
+        return self._get_by_href_impl(*a, **kw)
 
     def get_by_href_b(self, *a, **kw):
-        return self._get_by_href_impl(*a, **kw, side='b')
+        kw['side'] = 'b'
+        return self._get_by_href_impl(*a, **kw)
 
 
 class _Status(_StatusBase):
--- a/vdirsyncer/sync.py
+++ b/vdirsyncer/sync.py
@@ -176,9 +176,8 @@ class _SqliteStatus(_StatusBase):
         if self._is_latest_version():
             return
 
-        self._conn.execute('DROP TABLE IF EXISTS status')
-        self._conn.execute('DROP TABLE IF EXISTS new_status')
-        self._conn.execute('DROP TABLE IF EXISTS meta')
+        # If we ever bump the schema version, we will need a way to migrate
+        # data.
 
         self._conn.execute('''CREATE TABLE meta (
             "version" INTEGER PRIMARY KEY
--- a/vdirsyncer/sync.py
+++ b/vdirsyncer/sync.py
@@ -12,9 +12,12 @@ Yang: http://blog.ezyang.com/2012/08/how-offlineimap-works/
 Some modifications to it are explained in
 https://unterwaditzer.net/2016/sync-algorithm.html
 '''
+import abc
 import contextlib
+import functools
 import itertools
 import logging
+import sqlite3
 
 from . import exceptions
 from .utils import uniq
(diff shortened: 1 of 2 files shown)


untitaker:
I'd be glad if somebody could review this. This is a very large change.

vdirsyncer/sync.py
```diff
@@ -99,107 +102,328 @@ def to_ident_conflict(self, storage):
...
+            "href_b" TEXT NOT NULL,
+            "hash_a" TEXT NOT NULL,
+            "hash_b" TEXT NOT NULL,
+            "etag_a" TEXT NOT NULL,
+            "etag_b" TEXT NOT NULL
+        ); ''')
+        self._conn.execute('CREATE UNIQUE INDEX by_href_a ON status(href_a)')
+        self._conn.execute('CREATE UNIQUE INDEX by_href_b ON status(href_b)')
+
+        # We cannot add NOT NULL here because data is first fetched for the
```
hobarrera: Go get around this double-table thing, maybe we can split this into several tables:

* `entries`: has `ident`.
* `entry_hashes`: has columns:
  1. collection: a or b
  2. hash
  3. etag

This avoids any nulls, and you can read everything at once with a `JOIN`. Generally, I prefer to avoid `NULL` columns if the actual data they model cannot be null.

vdirsyncer/sync.py
```diff
@@ -99,107 +102,328 @@ def to_ident_conflict(self, storage):
...
+        try:
+            return bool(self._conn.execute(
+                'SELECT version FROM meta WHERE version = ?',
+                (self.SCHEMA_VERSION,)
+            ).fetchone())
+        except sqlite3.OperationalError:
+            return False
+
+    @contextlib.contextmanager
+    def transaction(self):
```
hobarrera: I'm not entirely sure what the method does, and I think that reflects a bit poorly on the naming. It clearly doesn't start a DB transaction though, which is what I expected.

vdirsyncer/sync.py
```diff
@@ -99,107 +102,328 @@ def to_ident_conflict(self, storage):
...
 
     def insert_ident_a(self, ident, props):
-        props_a, props_b = self._new_ident_to_props.get(ident, (None, None))
-        if props_a is not None:
-            raise _IdentAlreadyExists(old_href=props.href,
-                                      new_href=props_a['href'])
-        self._new_ident_to_props[ident] = props.to_status(), props_b
+        # FIXME: Super inefficient
+        old_props = self.get_new_a(ident)
+        if old_props is not None:
```
hobarrera: Can this be falsy without being `None`? Otherwise, just use `if not old_props`.

vdirsyncer/sync.py
```diff
@@ -99,107 +102,328 @@ def to_ident_conflict(self, storage):
...
+    def __init__(self, path):
+        self._path = path
+        self._conn = sqlite3.connect(path)
+        self._conn.row_factory = sqlite3.Row
+        self._c = None
+
+        if self._is_latest_version():
+            return
+
+        self._conn.execute('DROP TABLE IF EXISTS status')
```
untitaker: This is guarded by the early return three lines above, analogous to todoman's schema management.

hobarrera: Yes, I understand this. My point was, when the `SCHEMA_VERSION` gets pushed, you'll drop all status data, and can't do future syncs.

todoman merely does caching in sqlite, so we can afford to take this easy road and just rebuild everything. I don't think you can rebuild everything here.

vdirsyncer/sync.py
untitaker: I've thought about a few designs that involve more tables (because having `href_a` and `href_b` for the same datatype is ugly), but at least in my mind I couldn't find one that is significantly less complex at this stage.

I'm not sure how you'd have a `JOIN` between your two tables, they have no overlapping column.

The `NULL`s really need to be there for `new_status` because *while* the program is running, `NULL`s can occur. In the end the absence of `NULL` is still enforced when we copy all the data from `new_status` to `status`.

vdirsyncer/sync.py
untitaker: It was originally supposed to, and now it does again.

Originally I intended to do all changes within the `status` table, but I couldn't find a way to "defer" the NOT NULL constraint to the end of the transaction (something like "deferred NOT NULL constraints", similar to "deferred foreign key constraints"). There isn't even a way to change constraints of an existing table, so now there are two tables.

vdirsyncer/sync.py
untitaker: I'd really like this to stay explicit since types have been very confusing in this module in the past.

vdirsyncer/sync.py
hobarrera: > The NULLs really need to be there for new_status because while the program is running, NULLs can occur

I was actually thinking that you could only have *one* status table and not two status tables, since you can update (and persist) rows at once now in the new schema. Am I mistaken here?

Either way, it's probably out-of-scope for this PR and something that warrants a separate one.

vdirsyncer/sync.py
untitaker: Technically, if you sync directly before upgrading vdirsyncer, it should be fine that your status data gets rebuilt from scratch. It is very slow though because it will fetch all items from both sides, and store them all in RAM (same thing happens on first usage of vdirsyncer).

I didn't think about this at all though, so now I replaced all DROP TABLE statements with a placeholder comment.

vdirsyncer/sync.py
untitaker: During sync we have to access the old and new data simultaneously. Sqlite might support that with its transactions, but as explained further below we also need a way to temporarily disable NOT-NULL constraints.

vdirsyncer/sync.py
hobarrera: A placeholder or `NotImplemented` is enough for now. :+1:

vdirsyncer/sync.py
```diff
@@ -239,23 +238,31 @@ def insert_ident_a(self, ident, props):
             raise _IdentAlreadyExists(old_href=old_props.href,
                                       new_href=props.href)
 
-        self._c.execute('WITH new (ident, href_a, hash_a, etag_a) AS ( VALUES(?, ?, ?, ?) )'
-                        'INSERT OR REPLACE INTO new_status '
-                        'SELECT new.ident, new.href_a, old.href_b, new.hash_a, old.hash_b, new.etag_a, old.etag_b '
-                        'FROM new LEFT JOIN new_status AS old ON old.ident = new.ident',
-                        (ident, props.href, props.hash, props.etag))
+        self._c.execute(
```
hobarrera: Multiline strings might be better for this sort of thing.

vdirsyncer/sync.py
untitaker: Did you comment on the wrong section?

hobarrera: Yes, defeinitely. This was meant as a reply to the dropping tables thread, sorry.

vdirsyncer/sync.py
untitaker: It's within the line limit now.

vdirsyncer/sync.py
untitaker: BTW this doesn't silently fail, because `CREATE TABLE` will fail for existing schemas.

untitaker:
I have no idea what happened, but this appears to pass all tests locally.

untitaker:
master is now very slow because of all the copying.

PR #611, opened 2017-03-27
Switch to sqlite for sync status

### LINKED ISSUE #546: Switch status format to sqlite

### LINKED ISSUE #561: Lock pair during sync
I erroneously filed this at first as https://github.com/pimutils/khal/issues/590.

I just noticed that I can accidentally run sync from cron and from the terminal at the same time. I have not actually produced a problem from this, but unless synchronous write access is part of the design (which I doubt), this is going to burn someone.

Please lock a collection while it's being processed. It'd be great if this could be done per-pair, rather than for the entire run.

This is quite possibly related to #462, but also not the same, I think.

untitaker:
Fix #546 
Fix #561

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,9 @@ Version 0.16.0
 - Fix crash of Google storages when saving token file.
 - Make DAV discovery more RFC-conformant, see :ghpr:`585`.
 - Vdirsyncer is now tested against Xandikos, see :ghpr:`601`.
+- Statuses are now stored in a sqlite database. Users with really large
+  datasets should encounter performance improvements. This means that **sqlite3
+  is now a dependency of vdirsyncer**.
 
 Version 0.15.0
 ==============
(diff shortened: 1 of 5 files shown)


untitaker:
Weird error under OS X.

untitaker:
:pray:

PR #614, opened 2017-03-30
Initial etesync support

untitaker:
- [x] tests leave repo in dirty state


vdirsyncer/vobject.py
```diff
@@ -40,7 +40,7 @@ class Item(object):
     VCARD'''
 
     def __init__(self, raw):
-        assert isinstance(raw, str)
+        assert isinstance(raw, str), type(raw)
```
hobarrera: What exactly does `type(raw)` do here? I don't quite understand the syntax.

untitaker: `type(raw)` is the error message. It's not related but should be there IMO either way.

hobarrera: Oh, cool, didn't know the syntax.

geier:
IIRC, type(raw) will used as the AssertionError's parameter should
isinstance be Falseish

Quoting Hugo Osvaldo Barrera (2017-03-30 20:21:40)

Issue #228, opened 2015-07-07
Deprecate/discourage various forms of password storage?


untitaker:
E.g. `netrc` is just as insecure as storing the password in the config file, and I can't see any advantages in using it.

nblock:
One "advantage" is that the password only needs to be changed once, if a user/host is used multiple times.

untitaker:
Right, at the point where this matters, a real password manager should be used IMO.

geier:
If it isn't too much of a hassle to support those password storages I would leave them in but discourage from using them in the documentation. I believe there are still a lot of old school UNIXers out there who do use netrc files (netrc support was actually one of the first pull requests for pycarddav IIRC).

nblock:
I think netrc support is quite useful for unatteded usage such as cron jobs. At least, the barrier is easier than eg. using a custom command and gpg to get the password.

cyberic99:
What method woul you advise to use inside cron jobs, for instance?

hobarrera:
Personally, I prefer to keep the password *outside* the config file, since I keep my dotfiles for my machines in sync via git on a public repository (plus, there's the sharing factor). I know this is a rather common practice too and I'd hate to have passwords in config files.

untitaker:
I am not advocating for storing passwords inside config files, but storing them inside the .netrc is just as bad. Actually it's slightly worse since you now have a single insecure attack target. It is easy to write a small script that scans the user's PC for `.netrc` files. The same thing is not as easy to do with most password managers.

Ideally I'd like everybody to not store their credentials as plaintext files on their machines, but if you have a good reasons not to do that, you can still use the command-support from vdirsyncer to read from a plaintext file.

hobarrera:
> I am not advocating for storing passwords inside config files, but storing them inside the .netrc is just as bad. Actually it's slightly worse since you now have a single insecure attack target. It is easy to write a small script that scans the user's PC for .netrc files. The same thing is not as easy to do with most password managers.

Most of my dotfiles are public, and if each app kept it's own password in-config, I'd be unable to share these.
Having a single separate file makes this simpler. Not that I'm merely talking about convenience there; there's extra security with either approach.

> Ideally I'd like everybody to not store their credentials as plaintext files on their machines, but if you have a good reasons not to do that, you can still use the command-support from vdirsyncer to read from a plaintext file.

[...]

PR #622, opened 2017-04-15
Add 'netrc' strategy for keyring.

doronbehar:
Hi, I read [here](https://vdirsyncer.readthedocs.org/en/latest/keyring.html) that vdirsyncer enables to write username and password in a different file and use a dedicated script to fetch them. I must say I find it a little clunky.. Wouldn't it be much more elegant to use the `~/.netrc`?

### DIFF
--- a/docs/keyring.rst
+++ b/docs/keyring.rst
@@ -38,6 +38,23 @@ You can fetch the username as well::
 
 Or really any kind of parameter in a storage section.
 
+Accessing your netrc file
+-------------------------
+
+As shown above, you can use the ``command`` strategy to fetch your credentials
+from arbitrary sources. A very common usecase is to fetch your password from
+your ``~/.netrc`` file.
+
+the netrc_ Python module enables you to fetch passwords and logins from your
+``~/.netrc`` file.
+
+Basic usage::
+
+    username.fetch = ["netrc", "login", "example.com"]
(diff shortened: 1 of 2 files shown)


untitaker:
Hello,

First, thanks for your contribution. We've had a long discussion around netrc here: https://github.com/pimutils/vdirsyncer/issues/228

hobarrera:
I think you should just use `command` to fetch the password from `.netrc`. The above discussion explains why there's no explicit support.

For my part, I'd be much more willing to accept a documentation-only PR on how to use this approach (command that reads netrc), rather than merge this.

untitaker:
I agree, docs would be ok

On April 15, 2017 7:40:50 PM GMT+02:00, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>I think you should just use `command` to fetch the password from
>`.netrc`. The above discussion explains why there's no explicit
>support.
>
>For my part, I'd be much more willing to accept a documentation-only PR
>on how to use this approach (command that reads netrc), rather than
>merge this.
>
>-- 
>You are receiving this because you commented.
>https://github.com/pimutils/vdirsyncer/pull/622#issuecomment-294307654

-- 

doronbehar:
Great, thank you for your replay, I'll think of something.

PR #624, opened 2017-04-16
Add more instructions regarding `docs/keyring.rst`.

doronbehar:
Add minimal explanation for using external password manger named `pass`
to automatically safely and fetch the password.

### DIFF
--- a/docs/keyring.rst
+++ b/docs/keyring.rst
@@ -38,6 +38,16 @@ You can fetch the username as well::
 
 Or really any kind of parameter in a storage section.
 
+It might be recommended to use an external password manager such as ``pass``
+to fetch the password. (more details can be found in it's `home website`_).
+
+With ``pass``, you might find yourself writing something like this in your
+configuration file::
+
+    password.fetch = ["command", "/usr/bin/pass", "caldav"]
+
+.. _`home website`: https://www.passwordstore.org/
+
 Accessing the system keyring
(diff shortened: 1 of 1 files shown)


hobarrera:
Might want to mention that this won't work if running non-interactively (e.g.: via cron ).

Otherwise, looks good IMO.

docs/keyring.rst
```diff
@@ -38,6 +38,16 @@ You can fetch the username as well::
 
 Or really any kind of parameter in a storage section.
 
+It might be recommended to use an external password manager such as ``pass``
+to fetch the password. (more details can be found in it's `home website`_).
+
+With ``pass``, you might find yourself writing something like this in your
+configuration file::
+
+    password.fetch = ["command", "/usr/bin/pass", "caldav"]
```
untitaker: You can just use `pass` here instead of the full path.

docs/keyring.rst
```diff
@@ -38,6 +38,16 @@ You can fetch the username as well::
 
 Or really any kind of parameter in a storage section.
 
+It might be recommended to use an external password manager such as ``pass``
```
untitaker: "It might be recommended" sounds weird, I suggest something like "For example".

Also it might be better to introduce a new headline here.

untitaker:
@doronbehar any update on this?

PR #632, opened 2017-05-11
Use inode number in addition to mtime

### LINKED ISSUE #582: Use inode number and filesize for etag generation
For extra safety, the etag of a local file should not only contain the mtime, but also the filesize and inode number. I think all of those things are returned by `os.stat`. The modification needs to happen in `vdirsyncer.utils.get_etag_from_file`.

untitaker:
Fix #582


hobarrera:
I wonder if this could be a helper function in `atomic_writes`, and we could re-use it on all `pimutils` apps.

Including it there is kind of a stretch,but I feel like both things go hand in hand.

untitaker:
>Including it there is kind of a stretch,but I feel like both things go hand in hand.

Someday. Right now I don't think I have figured this out myself either.

PR #635, opened 2017-05-24
Add systemd.timer files and a relevant tutorial

hobarrera:
Include a `systemd.timer` file (the interval can be overridden, as mentioned in the docs).  
TBH, I'd suggest downstream packagers of systemd-based distributions include these files on installation
(activation needs to be done manually anyway).


untitaker:
Not sure, we also have a plist file we don't really package.

hobarrera:
Cool, well, this is ready-to-merge on my part.

untitaker:
Thanks!

PR #651, opened 2017-07-16
Use rust-vobject

untitaker:


untitaker:
This PR reimplements the vobject parsing in Rust, and embeds it using [snaek](https://github.com/mitsuhiko/snaek). Rust and Cargo have to be installed, but apart from that installation and usage of vdirsyncer should stay the same.

Open questions:

* What to port next? Vobject was an easy target. What are hotspots in the code which could benefit from performance improvements?
* How will this affect distro packaging?

untitaker:
cc @pimutils/packaging, preferences on how to integrate Rust modules into Python code?

This PR works as-is, although the tests might fail. I encourage you to run `python setup.py sdist` and see whether the tarball suits your needs.

hobarrera:
As far as publishing [to PyPI] goes, we'd have to start providing per-architecture wheels (plus an sdist). I guess we'll just target i686 and amd64 ourselves, since that should pick up the grand majority of the userbase, and they won't need to install rust if installing via `pip`.

If installing via a package manager, they won't need to install rust either, so the only users that'll need to change their flow *at all*, are those that manually install from source.

untitaker:
>I guess we'll just target i686 and amd64 ourselves

I would've published for x86_64 only tbh. Even ARM would be more relevant.

hobarrera:
Whoops, my above comment may be invalid (sorry, just read the `snaek` docs).

Does this mean that rust-vobject is a dependency that'd have to be installed separately on the users's system?

That would make installation via package manager the same. Packagers will need to package that lib (if absent) separately, but users installing via `pip` will suffer. Especially on OSX, apparently.

untitaker:
@hobarrera your comment is entirely correct. You don't need to install rust-vobject manually. It is automatically downloaded and compiled as part of `pip install`

hobarrera:
Cool. I like where this is going. 🙂

untitaker:
I'm somewhat interested in how this will go with distros that have unbundling policies since Rust's package manager compiles everything into one shared library that is then bundled with vdirsyncer. Unbundling this is quite complex.

hobarrera:
It'll be amusing to watch, without doubt.

Hopefully, they'll fail and package vdirsyncer as-is, which'll hopefully result in less bogus bug reports.

untitaker:
I don't think that's for the package maintainer to decide, and strongly doubt that they will just stop unbundling.

mathstuf:
Is rust-vobject on crates.io by itself? If so, unbundling is likely to happen unless you plan on making either separate tarballs available or keeping vdirsyncer and rust-vobject releases in sync (e.g., a release of rust-vobject can force a no-op release of vdirsyncer and vice versa) so that simple subpackages can be made from one source tarball.

untitaker:
Important releases of rust-vobject trigger a corresponding vdirsyncer release. Apart from that I don't see the point in doing the same vice-versa.

On 17 July 2017 21:53:33 GMT+02:00, Ben Boeckel <notifications@github.com> wrote:
>Is rust-vobject on crates.io by itself? If so, unbundling is likely to
>happen unless you plan on making either separate tarballs available or
>keeping vdirsyncer and rust-vobject releases in sync (e.g., a release
>of rust-vobject can force a no-op release of vdirsyncer and vice versa)
>so that simple subpackages can be made from one source tarball.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/pimutils/vdirsyncer/pull/651#issuecomment-315863848

-- 

mathstuf:
I guess that makes sense. I could see that being fine from a packager point-of-view. This is the canonical source for both projects rather than rust-vobject being a separate project just copied into here.

0-wiz-0:
Since this asked for explicit comments from packagers, here my point of view (for pkgsrc and/or NetBSD):

rust is not portable. The rust maintainers deliberately do not allow bootstrapping from a widely available language (like e.g. golang allows: build go-1.4.x with a C compiler, build newer go with that).

pkgsrc contains a package for rust (1.16), but that only supports NetBSD/amd64, not any of the other >30 architectures.

If you ask me: please do not depend on rust.

remilocherer:
On Tue, Jul 18, 2017 at 07:45:35PM +0000, Thomas Klausner wrote:
> Since this asked for explicit comments from packagers, here my point of view (for pkgsrc and/or NetBSD):
> 
> rust is not portable. The rust maintainers deliberately do not allow bootstrapping from a widely available language (like e.g. golang allows: build go-1.4.x with a C compiler, build newer go with that).
> 
> pkgsrc contains a package for rust (1.16), but that only supports NetBSD/amd64, not any of the other >30 architectures.
> 
> If you ask me: please do not depend on rust.
> 

I share this view (OpenBSD package maintiner).

mathstuf:
Pedancy: Rust is portable; the (only extant) compiler is not. There is work underway to make the compiler accessible from more architectures, but it does involve cross compiling (which, AFAIK, is easier with LLVM than GCC, but GCC has a large head start on the set of supported arches).

lfam:
Coming from GNU Guix, I have similar concerns regarding the availability of Rust on a variety of platforms. Worst case, we will only offer vdirsyncer on the easily supportable platforms.

untitaker:
As far as I understand Guix is Linux by default. Judging from that it's a different situation than with the BSDs.

I'll have to see how to make this work, but I definetly want to incorporate Rust at some point, in some way. Not like this, not right now though.

On 18 July 2017 23:08:07 GMT+02:00, lfam <notifications@github.com> wrote:
>Coming from GNU Guix, I have similar concerns regarding the
>availability of Rust on a variety of platforms. Worst case, we will
>only offer vdirsyncer on the easily supportable platforms.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/pimutils/vdirsyncer/pull/651#issuecomment-316198372

-- 

hobarrera:
I sympathise with the views provided regarding the BSDs, and the lack of architectures supported in the wild. I feel that it's probably best to postpone this, regrettably.

hobarrera:
Regarding [my comment](https://github.com/pimutils/vdirsyncer/pull/651#issuecomment-315837495) above:

Sorry if this sounded hostile to anyone, that was not the intention.

Distributions unbundling dependencies have caused both devs and users a lot of headache, and irreproducible issues, etc, so my point was more a way of saying _"I really hope they'd stop doing that"_. 

Nonetheless, my original narration is easily misinterpreted, so apologies to anyone who felt offended by it.

geier:
Interesting. Out of interest: do you know how you deal with recent firefox (which has started to ship some rust components AFAIK)? I would expect it to work at least on ARM64 (at least on linux).

untitaker:
@0-wiz-0 @remilocherer could you answer the question by @geier? I wonder about this too.

In either case I'll close this since it appears that this will cause too many problems for packaging.

0-wiz-0:
@untitaker : The plan is to stay with 52ESR and hope that an open source (no binary bootstrap necessary) compiler for rust comes into existence quickly enough. I'm not holding my breath :(

remilocherer:
@untitaker : OpenBSD -current provides now Firefox 55. But only for amd64 and i386. And of course it now has rust as build dependency (see [1]). From what I read Firefox was anyhow not really usable on sparc64 or macppc since modern browsers require so much resources. vdircyncer can still be useful on such platforms ;-)

[1] http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/www/mozilla-firefox/Makefile?rev=1.325&content-type=text/x-cvsweb-markup

untitaker:
@remilocherer I strongly doubt that all of vdirsyncer's dependencies work on those platforms. Most of them are certainly not properly tested against them.

(Apart from that I personally don't care about anything else than x86 and ARM)

0-wiz-0:
firefox in pkgsrc was updated to 55. Since rust is disabled, I guess the rust parts are missing.

On your last point, NetBSD is supported on lots of platforms. What makes you think vdirsyncer dependencies won't work on them?

untitaker:
No dependency of vdirsyncer has platform-specific code, but no dependency is
tested against those architectures either. If they're not tested they're
technically not supported and may just happen to work.

If you as a packager run the testsuite on those platforms it should probably be ok.

PR #653, opened 2017-07-19
Set defaults for CI and DETERMINISTIC_TESTS.

jelmer:
This makes it a bit easier to manually run individual tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -37,7 +37,7 @@ settings.register_profile("deterministic", settings(
     derandomize=True,
 ))
 
-if os.environ['DETERMINISTIC_TESTS'].lower() == 'true':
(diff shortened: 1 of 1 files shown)


untitaker:
BTW, I have a `make sh` shortcut which will initialize the correct environment variables. I wanted to avoid duplicating default values. Do you think that's a good solution that just has to be documented?

jelmer:
Ah, I didn't know about 'make sh'. Having that more visibly documented would be great (and is sufficient for what I was trying to do). 

(I've updated the default for DETERMINISTIC_TESTS to false, but please just close this pull request if you think "make sh" is sufficient)

untitaker:
I used to be more against this, but I think I'm fine with it now.

PR #730, opened 2018-04-01
Implement http storage in rust

untitaker:
second attempt. First attempt crashed with corrupted memory when I tried to deploy it to my server

TODO:
- [x] write tests for real this time (this is not "trivial")
- [x] fix the bug
- [ ] test on personal infra


untitaker:
So it turns out that the problem is somehow related to me instrumenting openssl 1.0 from both the embedded dylib and Python code. The problem goes away if I do *one* of:

- upgrade to openssl 1.1
- don't use the network code in Rust
- don't use the network code in Python

not really sure what to do here. It seems this is blocked until I port the other storages to rust.

mathstuf:
Note that this removed testing for `radicale`. Was that intended?

untitaker:
As part of this I got rid of a bunch of workarounds that only existed because of radicale

untitaker:
I genuinely don't know, but I don't want to block a release on it

mathstuf:
OK, sounds good.

Issue #765, opened 2018-09-11
google_calendar discover fails with missing import


djpohly:
Looks like this was probably caused by commit 04b6913, since that's where `prepare_verify` was removed.

vdirsyncer git ac45cf1, Python 3.7.0, Linux (Arch)

Config file:
````
[general]
status_path = "~/.local/share/vdirsyncer/status/"

[storage cal_local]
type = "filesystem"
path = "~/docs/calendar"
fileext = ".ics"

[storage cal_google]
type = "google_calendar"
token_file = "~/.local/share/vdirsyncer/token_google"
client_id = "redacted"
client_secret = "redacted"

[pair google]
a = "cal_local"
b = "cal_google"
conflict_resolution = "b wins"
collections = [["Google", "Google", "djpohly@gmail.com"]]
metadata = ["color"]
````

Output of `vdirsyncer -vdebug discover`:
````
error: Unknown error occurred: cannot import name 'prepare_verify' from 'vdirsyncer.http' (/usr/lib/python3.7/site-packages/vdirsyncer/http.py)
error: Use `-vdebug` to see the full traceback.
debug:   File "/usr/lib/python3.7/site-packages/vdirsyncer/cli/utils.py", line 342, in _worker
debug:     func(wq=self)
debug:   File "/usr/lib/python3.7/site-packages/vdirsyncer/cli/tasks.py", line 82, in discover_collections
debug:     rv = collections_for_pair(pair=pair, **kwargs)
debug:   File "/usr/lib/python3.7/site-packages/vdirsyncer/cli/discover.py", line 65, in collections_for_pair
debug:     b_discovered = _DiscoverResult(pair.config_b)
debug:   File "/usr/lib/python3.7/site-packages/vdirsyncer/cli/discover.py", line 129, in __init__
debug:     self._cls, _ = storage_class_from_config(config)
debug:   File "/usr/lib/python3.7/site-packages/vdirsyncer/cli/utils.py", line 250, in storage_class_from_config
debug:     cls = storage_names[storage_name]
debug:   File "/usr/lib/python3.7/site-packages/vdirsyncer/cli/utils.py", line 48, in __getitem__
debug:     mod = importlib.import_module(modname)
debug:   File "/usr/lib/python3.7/importlib/__init__.py", line 127, in import_module
debug:     return _bootstrap._gcd_import(name[level:], package, level)
debug:   File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
[...]

PR #788, opened 2019-02-20
fix google calendar sync

dpaetzel:
Dirty but simple fix for #765. Without this, syncing with Google calendar does not work any more.


WhyNotHugo:
Hi! We've rolled back master quite a bit (to the last table release, really).

Do you have issues with v0.16.7? Can you confirm that applying 7b636e8e40d69c495901f965b9c0686513659e44 to is also works for you?

Thanks.

dpaetzel:
Hi! All this PR did was reincluding `prepare_verify` (by copy-paste) which had been removed from the `http` module. It seems that in master, `prepare_verify` is again part of `http`.

However, master throws two kinds of errors for me:
```
cannot import name 'prepare_auth' from 'vdirsyncer.http'
```
and
```
critical: Failed to initialize uni_remote
critical:   - caldav storage doesn't take the parameters: verify_cert, auth_cert_password
```
which make 13 of 25 task fail (although this probably merits opening a new issue).

WhyNotHugo:
Look like your config is set up for the previous master (e.g.: what would have become v0.17).

We've rolled back to 0.16 and work is ongoing based on that, so you'll need to tweak your config file accordingly.

dpaetzel:
After deleting all local files and performing `discover`, the second error no longer occurs :heavy_check_mark: 

However, I still get this:
```
error: Unknown error occurred for google/Me: cannot import name 'prepare_auth' from 'vdirsyncer.http' (/home/david/Code/vdirsyncer/home/lib/python/vdirsyncer-0.16.8.dev68+g68ff37e.d20200614-py3.7.egg/vdirsyncer/http.py)
```
which doesn't seem config-related to me?

Issue #790, opened 2019-03-17
Maintenance


untitaker:
I'd like to hand this project over. I have multiple reasons for having stopped investing time in this project:

* I stopped using vdirsyncer and everything in pimutils, including using mutt since I started using OS X. I still self-host, but use GUI apps for many things.

* [Neverending bugreports about undebuggable servers. Inherent bugginess of everything DAV-related. I wrote about that already, but don't think anything will change there.](https://unterwaditzer.net/2015/kill-webdav.html)

* Horrible packaging story in Python, and worse packaging story for users who are used to installing everything from a distro's packaging manager. Downstream distributors don't provide a way to distribute test builds to end users. This triples the pain of working on a bugreport since most users need extra assistance in installing from sources, or simply wait for a release from me (and the corresponding downstream release) until they provide feedback on a bugfix.

  Larger projects provide their own Debian packages, so I tried that, but: 1.) it's too much work for one person 2.) there is fundamental distrust for things not found in Debian's official repos 3.) what about the other distros?

  Another thing I tried was rewriting the entire thing in Rust so I could distribute static binaries. I hoped that this would make installation on all platforms easier. Particularly on Windows it would make a lot of aspects of packaging easier, should I ever decide to support Windows. But that's where my energy just left me halfway through. Partly because the way I attempted the rewrite introduced a lot of friction, but mostly because I already started to loose interest in the project as a whole.

* This is the biggest one: The last point about packaging and the bugreport feedback loop is made worse by distros that have multiple stability channels. People report issues about "the latest vdirsyncer version" when running on Debian stretch, which is still on version 0.14. Those people happen to be the ones sending me direct emails because they are too paranoid to open a GitHub account. It's why you won't find my email address on my GitHub profile anymore.

[...]

PR #787, opened 2019-02-18
Upgrade Rust dependencies and fix build errors

Mange:
I cannot compile this on Arch Linux anymore (#778) because of a version of OpenSSL not supported by the current OpenSSL-dependent crates.

OpenSSL was depended on by hyper, which was depended on by reqwest. The latest version of reqwest pulls in a newer hyper, which has a lot of breaking changes because they removed most of the type-safe HTTP parts.
(I hear this is temporary)

Anyway, I've ported all build errors to the newer version of reqwest. Most of the changes are pretty much copied-pasted everywhere; I did not try to extract any patterns or utility functions as I didn't want to make any assumptions on how you want to design this.

Let me know if you want me to do more abstractions or reduce the code size.

PR #779, opened 2018-11-30
Testing: Use hypothesis setting load_profile to setup health check

marsam:
Since Hypothesis 3.72.0, assigning settings attributes throws an error.

See: https://github.com/HypothesisWorks/hypothesis/pull/1572

PR #798, opened 2019-05-16
Fix build and style

spk:
Hi @untitaker, i'm aware of #790 and tried to gather work from other PRs like https://github.com/pimutils/vdirsyncer/pull/787 and https://github.com/pimutils/vdirsyncer/pull/779 to have a green ci build and style hopefully; cheers


WhyNotHugo:
Thanks! We've however taken a different approach to getting back to a green master and something that can be easier to maintain, and this is no longer applicable.

Cheers!

PR #799, opened 2019-05-25
Follow advice from setuptools_scm how to use it.

0-wiz-0:
See https://github.com/pypa/setuptools_scm#usage-from-sphinx

(The build broke for me with setuptools_scm 3.3.3)


WhyNotHugo:
Thanks! Merged via b1214cd.

Issue #819, opened 2020-05-08
Please release a version with stated Python 3.7 support


equaeghe:
* Your vdirsyncer version: _0.16.7 _
* Your Python version: _moving from 3.6 to 3.7_
* Your operating system: Gentoo Linux
---
Gentoo is moving from 3.6 to 3.7 as the default Python. vdirsyncer 0.16.7 states support up to and including Python 3.6:
https://github.com/pimutils/vdirsyncer/blob/7a92aa20b1910eb5cc3c451a9a2fa62e76c80141/setup.py#L126-L129
Unless vdirsyncer is indeed incompatible with Python 3.7, please release a version with Python 3.7 added here. Then downstream (Gentoo) can ‘allow’ this Python version as well.

(Of course, in case vdirsyncer is compatible with Python 3.8 as well, please add that version too.)

If there are known issues with Python 3.7 encountered by users, please refer to them in this issue, so that we can get an overview. Please do not describe unreported compatibility issues here, add them as a separate issue first.

Users that run vdirsyncer with Python 3.7+, please add a +1/thumbs up to this post. Users that run it with Python 3.8, please additionally add a Hooray cone. This allows us to get the impression of compatibility in practice.

WhyNotHugo:
I've been using vdirsyncer with Python 3.8 for ages now. I'll update CI to run with Python 3.7 and 3.8, if all tests pass, I guess we can officially support these.

WhyNotHugo:
Hmm... master/0.17 seems quite broken, and it's slightly beyond me to fix it.

Maybe a point release for 0.16 might make more sense.

equaeghe:
> Maybe a point release for 0.16 might make more sense.

Yes, that would be ideal!

WhyNotHugo:
0.16.8 has been release. It support Python 3.5 thru 3.8.

PR #820, opened 2020-05-08
Update supported Pythons

WhyNotHugo:
- Drop Python 3.4
- Support Python 3.7 and 3.8.
- Update references to these things.

See #819


WhyNotHugo:
Hmm... CI seems to fail building the rust openssl module, but it's not clear to me why.
I can't reproduce locally since failing fails at an earlier stage for me:
```
       Compiling url v1.7.0
       Compiling generic-array v0.9.0
    error[E0713]: borrow may still be in use when destructor runs
       --> /home/hugo/.cache/cargo/registry/src/github.com-1ecc6299db9ec823/url-1.7.0/src/form_urlencoded.rs:261:40
        |
    259 | impl<'a> Target for ::UrlQuery<'a> {
        |      -- lifetime `'a` defined here
    260 |     fn as_mut_string(&mut self) -> &mut String { &mut self.url.serialization }
    261 |     fn finish(self) -> &'a mut ::Url { self.url }
        |                                        ^^^^^^^^ - here, drop of `self` needs exclusive access to `*self.url`, because the type `UrlQuery<'_>` implements the `Drop` trait
        |                                        |
        |                                        returning this value requires that `*self.url` is borrowed for `'a`

    error: aborting due to previous error

    For more information about this error, try `rustc --explain E0713`.
    error: could not compile `url`.

    To learn more, run the command again with --verbose.
    warning: build failed, waiting for other jobs to finish...
    error: build failed
```

WhyNotHugo:
@untitaker I think you're the expert on rust here, any hints?

untitaker:
@WhyNotHugo I don't have a lot of time to look into this but I assume either pkg-config or even openssl is entirely missing from the circleci container. Maybe they changed something there.

Issue #821, opened 2020-06-05
password.fetch depends on pwd


Witcher01:
This issue is about the `password.fetch` field in the configuration file.  
It seems that this depends upon the present working directory of the user calling vdirsyncer.  
With the current configuration, calling vdirsyncer in my $HOME results in the message:
```
gpg: can't open '.vdirsyncer_posteo.gpg': No such file or directory
gpg: decrypt_message failed: No such file or directory
```
Calling it from `$XDG_CONFIG_HOME/vdirsyncer/` works just fine. But that is to be expected when giving a relative path as the argument to gpg.

I first tried setting a full path in `password.fetch` as an argument to gpg, but that doesn't seem to work with either `~/.config/vdirsyncer/...` nor `$HOME/.config/vdirsyncer/...` nor `$XDG_CONFIG_HOME/vdirsyncer/...`. It seems like the path given as an argument is alway interpreted as a relative path.  
Trying to execute gpg with the same path argument (full path) results in correct execution, so I highly suspect this is an issue with vdirsyncer passing/interpreting the argument.

* Your vdirsyncer version
`vdirsyncer, version 0.16.7`
* Your Python version
`Python 3.8.3`
* Your operating system
`Linux 5.4.39-1-lts GNU/Linux (Arch Linux)`
* Your config file
relevant line:
`password.fetch = ["command", "gpg", "-dq", ".vdirsyncer_posteo.gpg"]`

WhyNotHugo:
Fixed via #839.

PR #839, opened 2020-07-24
fixed password.fetch option not using environment

Witcher01:
the '_strategy_command' in 'fetchparams.cli' did not expand the path of
every argument, but only the first one (being the command to be
executed).
i fixed the '_strategy_command' function to apply the 'expand_path'
function to every member of the commands list.
this is in reply to issue #821


WhyNotHugo:
Merged via c254b4ad1d86745fc0389eea02ea687d48c4feee. Thanks!

PR #846, opened 2020-10-04
contrib/systemd: implement service timeout

pfactum:
Sometimes while doing a suspend/resume cycle vdirsyncer can hang in the
middle of sync, and it never bails out afterwards.

Implement a systemd service change that works around this:

* use `simple` (a default) type for `RuntimeMaxSec` to be effective
* actually set `RuntimeMaxSec` to 30 seconds
* trigger service restart on failure

Signed-off-by: Oleksandr Natalenko <oleksandr@natalenko.name>

### DIFF
--- a/contrib/vdirsyncer.service
+++ b/contrib/vdirsyncer.service
@@ -4,4 +4,5 @@ Documentation=https://vdirsyncer.readthedocs.org/
 
 [Service]
 ExecStart=/usr/bin/vdirsyncer sync
-Type=oneshot
+RuntimeMaxSec=30s
+Restart=on-failure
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Seems to make sense. I take it that due to the restart condition it behaves "kinda" like a `oneshot`.

I'd rather extend the duration quite a bit though -- maybe up to 3 minutes? I expect that some people may have mixes of many servers and/or slow connections, and I'd rather not break for them. There little harm in waiting 3 minutes anyway.

pfactum:
@WhyNotHugo please check f79647b for an extension to 3 mins.

WhyNotHugo:
Perfect, thanks!

PR #868, opened 2021-02-28
cli/discover: remove local collections if the remote collection is deleted

dilyanpalauzov:
This works when the destination backend is 'filesystem' and the source is CalDAV-calendar-home-set.

### DIFF
--- a/vdirsyncer/cli/__init__.py
+++ b/vdirsyncer/cli/__init__.py
@@ -1,10 +1,13 @@
 import functools
 import logging
+import os
+import shutil
 import sys
 
 import click
 import click_log
 
+from ..utils import expand_path
 from .. import __version__
 from .. import BUGTRACKER_HOME
 
@@ -198,6 +201,9 @@ def discover(ctx, pairs, max_workers, list):
     config = ctx.config
     wq = WorkerQueue(max_workers)
 
+    pairs_a = set()
+    pairs_b = set()
+    pairs_b_path = {}
     with wq.join():
         for pair_name in (pairs or config.pairs):
             pair = config.get_pair(pair_name)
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
I'm not a fan of doing this by default.

I've several times deleted a remote collection with the intent of pushing my local copy elsewhere. If my local copy were auto-deleted, I would have lost it permanently.

I'm sure there's other scenarios where it might lead to people inadvertently losing data. I don't want people finding out about this feature the hard way :sweat_smile: 

I'd be okay with this with any of these changes a dedicated flag to enable / disable the behaviour. For example, something like:

```dosini
[pair calendars]
# ...
prune_collections = true
```

dilyanpalauzov:
I am utterly fine in anyway this is implemented.  I provided my way of doing it.  I have also
```diff
diff --git a/vdirsyncer/cli/utils.py b/vdirsyncer/cli/utils.py
--- a/vdirsyncer/cli/utils.py
+++ b/vdirsyncer/cli/utils.py
@@ -404,7 +404,7 @@ def handle_collection_not_found(config, collection, e=None):
                        .format(f'{e}\n' if e else '',
                                json.dumps(collection), storage_name))
 
-    if click.confirm('Should vdirsyncer attempt to create it?'):
+    if True or click.confirm('Should vdirsyncer attempt to create it?'):
         storage_type = config['type']
         cls, config = storage_class_from_config(config)
         config['collection'] = collection

```
to create automatically Calendars, once they are created on the server.  I want to have a system that periodically syncs and gets the state on the server (incl. creating new calendars and deleting old ones).  For implicit calendar creation, the name `prune_collection` is not good, if both options will get the same name.

dilyanpalauzov:
Moved to https://github.com/pimutils/vdirsyncer/pull/869.

PR #869, opened 2021-03-07
cli/discover: remove/add local collections if the remote collection is deleted/created

dilyanpalauzov:
This works when the destination backend is 'filesystem'.

-- add a new parameter to storage section:
   implicit = ["create", "delete"]

Changes cli/utils.py:save_status(): when data is None, remove the underlaying file.


vdirsyncer/cli/discover.py
```diff
@@ -80,6 +80,21 @@ def collections_for_pair(status_path, pair, from_cache=True,
         get_b_discovered=b_discovered.get_self,
         _handle_collection_not_found=handle_collection_not_found
     ))
+    only_in_b = set(b_discovered.get_self().keys()) - set(a_discovered.get_self().keys())
```
WhyNotHugo, line 83: If I understand correctly, this block will delete from `b` any collections absent from `a`, right?

That makes sense if settings include:

```dosini
collections = ["from a"]
```

However, if settings are as follows, then you'd be deleting from the wrong side, right?


```dosini
collections = ["from b"]
```

dilyanpalauzov: That is correct.  I uploaded new code that considers "from a" and "from b".  I do not understand all the advanced ways to set collections.

dilyanpalauzov:
The tests fail with
```
    raise DeadlineExceeded(runtime, self.settings.deadline)
        hypothesis.errors.DeadlineExceeded: Test took 227.47ms, which exceeds the deadline of 200.00ms
```

dilyanpalauzov:
Will this get eventually integrated upstream?

WhyNotHugo:
@dilyanpalauzov Yup, sorry, just been busy, but the change makes sense. I'll deal with conflicts when I get a change :)

dilyanpalauzov:
Please elaborate what means “when I get a change”?

WhyNotHugo:
Sorry, I meant "why I get a chance'. Meaning when I get a bit of free time to do the conflict resolution.

WhyNotHugo:
I've rebased your branch onto `master` and resolved all conflicts.

I'm a bit uncertain about how this works when you have:

- storage A with `implicit = "create"`
- storage B with `implicit = "delete"`

If there's a collection present in B, but missing in A, should it be deleted from B, or created in A?

dilyanpalauzov:
I have not thought on this case.  I would say, the collection shall be created.

Rationale: I do not know why this combination is good for.  Preventing data deletion in this unclear use-case is a good approach, as long as nobody proposes something better.

dilyanpalauzov:
Let me know if there is anything from my side, which shall be done for this PR.  I do not understand the code of vdirsyncer very good.

dilyanpalauzov:
>    • storage A with implicit = "create"
>    • storage B with implicit = "delete"
> 
> If there's a collection present in B, but missing in A, should it be deleted from B, or created in A?

How about writing down,that the behaviour in this case is not defined.

samsonjs:
Is there anything I can do to help with this one? I'd love to have this feature as well.

samsonjs:
Perhaps only implicit creation should be allowed? That way there's never any accidental data loss, which is far worse than a zombie collection coming back from the dead. And there's no ambiguity about what to do in this scenario:

> I'm a bit uncertain about how this works when you have:
>
> storage A with `implicit = "create"`
> storage B with `implicit = "delete"`
>
> If there's a collection present in B, but missing in A, should it be deleted from B, or created in A?

Selfishly, implicit creation is the only part of this that I actually need so if eliminating implicit deletion helps move this forward at all here then that suits me just fine 😄

dilyanpalauzov:
I need also implicit deletion.

samsonjs:
Fair enough. What about using `conflict_resolution` for this too? In the example scenario above there would be an error if none is specified, or if the config has `"b wins"` then the collection is deleted from A. I'm not sure if there's a good way use the diff command for this one though so maybe it could also produce an error when a command is configured.

I think that behaviour is actually pretty intuitive and most people would expect it to work that way.

dilyanpalauzov:
This pull request from me is stalled.  I do use the changes on my system and it does work exactly as I want.  I will appreciate any progress on this, but I am not going to modify the proposed changes.

PR #877, opened 2021-04-07
Some code simplifications with the return statement

dilyanpalauzov:


vdirsyncer/storage/dav.py
```diff
@@ -138,9 +138,7 @@ def _fuzzy_matches_mimetype(strict, weak):
         return True
 
     mediatype, subtype = strict.split('/')
-    if subtype in weak:
-        return True
-    return False
+    return subtype in weak
```
WhyNotHugo, line 141: This is a much welcome cleanup though.

vdirsyncer/http.py
```diff
@@ -37,10 +37,10 @@ def prepare_auth(auth, username, password):
     if username and password:
         if auth == 'basic' or auth is None:
             return (username, password)
-        elif auth == 'digest':
+        if auth == 'digest':
             from requests.auth import HTTPDigestAuth
             return HTTPDigestAuth(username, password)
-        elif auth == 'guess':
+        if auth == 'guess':
```
WhyNotHugo, line 43: I don't really like this change.

It makes it far harder to figure out the flow. At a glance, I'd get the impression that this code run `if auth == 'guess'`, but there's actually a return nested further above.

The `elif` makes it a bit more obvious that these are mutually exclusive code-paths.

vdirsyncer/cli/utils.py
```diff
@@ -181,7 +181,7 @@ def get_status_path(base_path, pair, collection=None, data_type=None):
 def load_status(base_path, pair, collection=None, data_type=None):
     path = get_status_path(base_path, pair, collection, data_type)
     if not os.path.exists(path):
-        return None
+        return
```
WhyNotHugo, line 184: :+1:

vdirsyncer/http.py
dilyanpalauzov: In all programming languages, writing `else` after `return` has no added value.

WhyNotHugo: It helps readability, IMO.

WhyNotHugo:
Rebased and applied. Thanks.

vdirsyncer/cli/discover.py
```diff
@@ -59,21 +59,20 @@ async def collections_for_pair(
     cache_key = _get_collections_cache_key(pair)
     if from_cache:
         rv = load_status(status_path, pair.name, data_type="collections")
-        if rv.get("cache_key", None) == cache_key:
+        if rv and rv.get("cache_key", None) == cache_key:
```
dilyanpalauzov, line 62: Is this change form me and is it necessary?  The other changes here do not modify the program logic, that is they preserve all the bugs.  But this change modifies the program logic.

WhyNotHugo: This change was from your original commit.

When I reviewed this earlier the change made sense to me because of the following `elif rv`. I realise now that it is unnecessary. It should not change the logic, assuming that `rv` is never `None`.

PR #889, opened 2021-05-25
Implement more flexibility for storage/filesystem

### LINKED ISSUE #881: Enhancement: storage type or config to be used with Kontact's directory handling
https://kontact.kde.org/ has a power calendar component called https://kontact.kde.org/components/korganizer.html

Modern versions can use CalDAV directly, but elder versions (kontact enterprise35) still in use could not, but
offer saving of calendars in single files, much like the `vdir storage format`.  It would be useful for us to use
vdirsyncer to make this old kontact enterprise35 to work with CalDAV servers.

Because of the similiarity only small changes would be necessary. Two differences already found are
 *  that the kontact-e35-dir format uses no suffixes
 * and a trailing `~` to mark files which are to be deleted and are not to be shown anymore.

bernhardreiter:
* add parameter `#fileignoreext = ".tmp"` to storage/filesystem code, tests and docs.
 * add hint to docs that vdir compatibility is only given with setting `fileext` properly

resolves #881

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -408,6 +408,7 @@ Local
       fileext = "..."
       #encoding = "utf-8"
       #post_hook = null
+      #fileextignore = ".tmp"
 
     Can be used with `khal <http://lostpackets.de/khal/>`_. See :doc:`vdir` for
     a more formal description of the format.
@@ -426,6 +427,8 @@ Local
     :param post_hook: A command to call for each item creation and
         modification. The command will be called with the path of the
         new/updated file.
+    :param fileextignore: The file extention to ignore,
+        the default is ``.tmp``.
 
 .. storage:: singlefile
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -422,13 +422,15 @@ Local
     :param fileext: The file extension to use (e.g. ``.txt``). Contained in the
         href, so if you change the file extension after a sync, this will
         trigger a re-download of everything (but *should* not cause data-loss
-        of any kind).
+        of any kind). To be compatible with the ``vset to the empty string`` format you have
+        to either use ``.vcf`` or ``.ics``. Note that metasync won't work
+        if you use an empty string here.
     :param encoding: File encoding for items, both content and filename.
     :param post_hook: A command to call for each item creation and
         modification. The command will be called with the path of the
         new/updated file.
-    :param fileeignoreext: The file extention to ignore,
-        the default is ``.tmp``.
+    :param fileeignoreext: The file extention to ignore. It is only useful
+        if fileext is set to the empty string. The default is ``.tmp``.
 
 .. storage:: singlefile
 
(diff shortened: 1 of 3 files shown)


bernhardreiter:
Commits can be squashed

bernhardreiter:
Any idea why the two tests fail?

docs/config.rst
```diff
@@ -421,11 +422,15 @@ Local
     :param fileext: The file extension to use (e.g. ``.txt``). Contained in the
         href, so if you change the file extension after a sync, this will
         trigger a re-download of everything (but *should* not cause data-loss
-        of any kind).
+        of any kind). To be compatible with the ``vset to the empty string`` format you have
```
WhyNotHugo, line 425: I think there's something wrong in this sentence.

vdirsyncer/storage/filesystem.py
```diff
@@ -22,13 +22,15 @@ class FilesystemStorage(Storage):
     storage_name = "filesystem"
     _repr_attributes = ("path",)
 
-    def __init__(self, path, fileext, encoding="utf-8", post_hook=None, **kwargs):
+    def __init__(self, path, fileext,
+                 encoding="utf-8", post_hook=None, fileignoreext=".tmp", **kwargs):
```
WhyNotHugo, line 26: Please run `black` which will handle auto-formatting everything for you :)

WhyNotHugo (commented):
Thanks! And thanks for including tests too!

Note sure why that test is failing. I've pushed a tiny change to a separate branch to get some better debug output.

Still running here: https://builds.sr.ht/~whynothugo/job/513264

WhyNotHugo:
Ugh, the CI failure is because of https://github.com/click-contrib/click-threading/pull/5

docs/config.rst
bernhardreiter: Fixed, thanks for noticing this "paste and copy" error.

vdirsyncer/storage/filesystem.py
bernhardreiter: Ran `black` (Personally I prefer hand-formatted code, as it is more expressive.) Thanks for the feedback!

WhyNotHugo:
Sorry for the delays, CI was broken with unrelated issues and that sucked up time.

I'm curious on this item though: why are you not using `fileext`? It would seem that an ignore would be unnecessary if specifying that.

bernhardreiter:
> why are you not using fileext? It would seem that an ignore would be unnecessary if specifying that.

Because of the format that Kontact (e35) uses, where regular calendar entries do not have a suffix. See #881 for details.

`fileext` has to be set to the empty string to recognize the files and then all the backup files will be found as well.

WhyNotHugo:
Thanks for reminding me, I'd lost track of that details.

WhyNotHugo:
Merged, thanks!

PR #893, opened 2021-06-11
Add contrib script to resolve conflicts

bernhardreiter:
Asks the user when two iCalender objects conflict during a sync,
which one to take. Documentation within the file itself.

### DIFF
--- a/contrib/conflict_resolution/resolve_interactively.py
+++ b/contrib/conflict_resolution/resolve_interactively.py
@@ -12,7 +12,7 @@ Usage:
   conflict_resolution = ["command", "/home/bern/vdirsyncer/resolve_interactively.py"]
 
 This file is Free Software under the following license:
-SPDX-License-Identifier: Apache-2.0
+SPDX-License-Identifier: BSD-3-Clause
 SPDX-FileCopyrightText: 2021 Intevation GmbH <https://intevation.de>
 Author: <bernhard.reiter@intevation.de>
 """
--- a/contrib/conflict_resolution/resolve_interactively.py
+++ b/contrib/conflict_resolution/resolve_interactively.py
@@ -16,16 +16,17 @@ SPDX-License-Identifier: BSD-3-Clause
 SPDX-FileCopyrightText: 2021 Intevation GmbH <https://intevation.de>
 Author: <bernhard.reiter@intevation.de>
 """
-from pathlib import Path
 import re
 import subprocess
 import sys
+from pathlib import Path
 
 KDIALOG = "/usr/bin/kdialog"
 
 SUMMARY_PATTERN = re.compile("^(SUMMARY:.*)$", re.MULTILINE)
 
-def get_summary(icalendar_text:str):
+
+def get_summary(icalendar_text: str):
     """Get the first SUMMARY: line from an iCalendar text.
 
     Do not care about the line being continued.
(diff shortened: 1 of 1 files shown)


contrib/conflict_resolution/resolve_interactively.py
```diff
@@ -0,0 +1,68 @@
...
+
+Depends on python>3.5 and KDialog.
+
+Usage:
+  Ensure the file executable and use it in the vdirsyncer.conf file, e.g.
+
+  conflict_resolution = ["command", "/home/bern/vdirsyncer/resolve_interactively.py"]
+
+This file is Free Software under the following license:
+SPDX-License-Identifier: Apache-2.0
```
WhyNotHugo, line 15: `vdirsyncer` is BSD-licensed, so including a file under this licence would be problematic since the source thing can't trivially be redistributed together.

contrib/conflict_resolution/resolve_interactively.py
```diff
@@ -0,0 +1,68 @@
...
+SPDX-FileCopyrightText: 2021 Intevation GmbH <https://intevation.de>
+Author: <bernhard.reiter@intevation.de>
+"""
+from pathlib import Path
+import re
+import subprocess
+import sys
+
+KDIALOG = "/usr/bin/kdialog"
```
WhyNotHugo, line 24: The hardcode is fine; I'm sure someone else interested in using the script will come along and make it parametrizeable.

contrib/conflict_resolution/resolve_interactively.py
```diff
@@ -0,0 +1,68 @@
...
+        return # shall lead to items not changed, because not copied
+
+    if r.returncode == 0:
+        # we want to take the first item, so overwrite the second
+        ical2_filename.write_text(ical1)
+    else: # r.returncode == 1, we want the second item, so overwrite the first
+        ical1_filename.write_text(ical2)
+
+if len(sys.argv) != 3:
+    sys.stdout.write(__doc__)
```
WhyNotHugo, line 66: `__doc__` doesn't actually give better hints on what parameters are necessary.

WhyNotHugo (changes requested):

contrib/conflict_resolution/resolve_interactively.py
bernhardreiter: In my view It does, as it describes how this is to be used in the config of vdirsyncer. (And then vdirsyncer gets the parameters right and describes them.) This is not to be called on its own and the __doc__ says so.

Okay for you, or should I be more explicit?

contrib/conflict_resolution/resolve_interactively.py
bernhardreiter: There are many Free Software sources that are distributed with parts that are under different licenses. Usually that is easily done, and it is under `contrib/`, so I haven't given it much thought. If you fell it is important I can change the license (and first have to check which of the "BSD"-Licenses is actually used and find its SPDX-License-Identifier.. ;) (Note the BSD-Licenses are not ideal, if not for the problem that they mix licensing text and copyright owner notes.)

contrib/conflict_resolution/resolve_interactively.py
bernhardreiter: Note that the interface to kdialog is implemented in the whole script, like parameter names and return codes, but it serves as an example how to do it, even with a different interactive application.

contrib/conflict_resolution/resolve_interactively.py
WhyNotHugo: It's the three-clause BSD licence. `SPDX short identifier: BSD-3-Clause`

I'd rather not include things in the repository that are under a different licence, it just make redistribution more complex (redistributors need to keep in that different bits have different licences).

I can include this under an Apache license, but that means that the source can never be reused or merged into `vdirsyncer` in any way, which creates a weird scenario for contributors too.

contrib/conflict_resolution/resolve_interactively.py
WhyNotHugo: That's a fair point, I guess it's never meant to be used interactively anyway.

WhyNotHugo:
Please run the `pre-commit` hooks to address the style-check issues.

contrib/conflict_resolution/resolve_interactively.py
bernhardreiter: License changed accordingly.

bernhardreiter:
> Please run the pre-commit hooks to address the style-check issues.

Ran flake8, black and reorder_python_imports. The code is less readable at the point where the command line is assembled, but the codestyle is more consistent.

Explanation: Setting up the pre-commit hooks the half-automatic way it was suggested in the repo was too magic for me, I would have had to read through it in details, thus I've used single modules.

WhyNotHugo:
> The code is less readable at the point where the command line is assembled, but the codestyle is more consistent.

I agree that it's less readable. I'll try to tweak something within the rules.

> Explanation: Setting up the pre-commit hooks the half-automatic way it was suggested in the repo was too magic for me, I would have had to read through it in details, thus I've used single modules.

No worries, `pre-commit` is there merely for convenience, and it's also important to use all the tools manually at least once.

FWIW, what pre-commit does is generate virtualenvs with the repositories in [`.pre-commit-config.yaml`](https://github.com/pimutils/vdirsyncer/blob/e467809b/.pre-commit-config.yaml). It then runs a command described in them, passing as parameters _only_ files that have changed in the current commit.

---

CI failures are unrelated here. Thanks for your contribution and thanks for your patience with delays :sweat_smile:

WhyNotHugo:
Maybe e16f83c1c23e874ad316110c67e4aed74bd662bb ?

bernhardreiter:
> Maybe [e16f83c](https://github.com/pimutils/vdirsyncer/commit/e16f83c1c23e874ad316110c67e4aed74bd662bb) ?

Looks good to me. (Did not run the code.)
(My remark about the code styling was more general.)

Thanks for merging and maintaining vdirsyncer!

PR #906, opened 2021-06-17
Initial asyncio support

WhyNotHugo:
So far, all storage test pass with xandikos (a few more with radicale) with the `dav` storages. Haven't gotten to the `google` storage.

All `system` and `unit` tests pass. Manually experimenting with `sync`, `discover` and `metasync` the basics all seem intact, but this still needs more thorough testing.

Non-dav storages have not really been ported, they just "work with the new async interface", but don't really operate async, so don't really have any performance changes.

For dav storages, preliminary results show that this is much faster than multi-threaded support, but there's still a lot of work to be done to really unlock all the benefits we can get from using `asyncio`. There's some reduction in code complexity by trimming out the threading support too.

I'm not 100% happy about passing a `TCPConnector` around so much, but it makes limiting the maximum amount of connections very easily at a global level.

I'll likely need some special wrapper to limit the amount of open files in a similar way (with a limit and a queue).

### Missing

- [x] `google` storage to `asyncio`. Still need to find a library for aiohttp + oauth.
- [x] `google` storage testing. There's no tests for that, so need to set some up to make sure this still works.
- [ ] `etesync` is still tied to run as a module rather than standalone, and won't work with our dependencies.

### Other changes

- Only `sha256` fingerprints for certificates are not supported. `sha1` and `md5` are insecure and have been dropped.

### DIFF
--- a/vdirsyncer/cli/__init__.py
+++ b/vdirsyncer/cli/__init__.py
@@ -65,33 +65,6 @@ def app(ctx, config):
 main = app
 
 
-def max_workers_callback(ctx, param, value):
-    if value == 0 and logging.getLogger("vdirsyncer").level == logging.DEBUG:
-        value = 1
-
-    cli_logger.debug(f"Using {value} maximal workers.")
-    return value
-
-
-def max_workers_option(default=0):
-    help = "Use at most this many connections. "
-    if default == 0:
-        help += (
-            'The default is 0, which means "as many as necessary". '
-            "With -vdebug enabled, the default is 1."
-        )
-    else:
-        help += f"The default is {default}."
-
-    return click.option(
-        "--max-workers",
-        default=default,
-        type=click.IntRange(min=0, max=None),
-        callback=max_workers_callback,
-        help=help,
-    )
-
-
 def collections_arg_callback(ctx, param, value):
     """
     Expand the various CLI shortforms ("pair, pair/collection") to an iterable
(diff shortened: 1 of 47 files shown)


vdirsyncer/cli/__init__.py
```diff
@@ -151,53 +125,56 @@ def sync(ctx, collections, force_delete, max_workers):
...
+                    collection=collection,
+                    general=config,
                     force_delete=force_delete,
-                    callback=sync_collection,
+                    connector=conn,
                 )
-            )
-            wq.spawn_worker()
+
+    loop = asyncio.get_event_loop()
```
graingert, line 145: get_event_loop is deprecated, use asyncio.run

WhyNotHugo: Good catch, thanks!

vdirsyncer/storage/base.py
```diff
@@ -92,6 +92,7 @@ class to obtain a storage instance pointing to this collection. It
             from the last segment of a URL or filesystem path.
 
         """
+        yield  # Needs to be an async generator
```
graingert, line 95:
```suggestion
        if false:
            yield  # Needs to be an async generator
```

WhyNotHugo: Ah, yes, that's better.

graingert: You can also put the yield after the raise

WhyNotHugo: I expect mypy would complain that it's unreachable code.

I'm surprised flake doesn't pick that up.

PR #907, opened 2021-06-18
Fix pip downloading all versions of pip on CI

WhyNotHugo:
Due to the new resolver, lots of different versions were being
downloaded and filled up /tmp.

Unsure if it's an upstream issue or not, but just be eager for now.


WhyNotHugo:
This was a bug in `pip` and not on our end.

The offending version of `pip` has since been yanked from upstream.

PR #871, opened 2021-03-08
metasync: use None as no-value and delete missing values on syncing

dilyanpalauzov:
- change the interface of Storage.get_meta() and .set_meta()  to use '' as the empty value and None as missing value.

- When a property is missing (e.g calendar-color was removed) in the filesystem storage delete the 'color' file, and in the WebDAV storage issue propertyupdate/remove call.

- remove the property from [status]/[pair]/[collection].metadata

I have adjusted, but not run the test suite.  On the other side I tested on real system: source is CalDAV - destination is filesystem; and source is CalDAV, destination is CalDAV (another account)

PR #920, opened 2021-07-23
metasync: use None as no-value and delete missing values on syncing

WhyNotHugo:
This is #871 rebased onto the current `master`.

Closes #871


WhyNotHugo:
It seems xandikos does not support deleting the property, I've reported this in https://github.com/jelmer/xandikos/issues/137.

PR #942, opened 2021-10-05
Drop etesync

### LINKED ISSUE #638: EteSync: implement metasync
The EteSync storage backend doesn't implement the "metasync" directive.

When trying to run it errors with: "This storage does not support metadata."

EteSync itself does however support metadata.

vdirsyncer version: current master

### LINKED ISSUE #827: Run etesync tests via docker
Currently we install these in the same container as vdirsyncer and use `wsgi_intercept`.

This makes creating the test environment very complicated, and since they both share the same, it's quite dirty. Try to figure out how to run etesync in a container, and test it a bit more like other servers.

### LINKED ISSUE #875: Re-enable ETESYNC_TESTS
When redoing CI, and fixing a lot of piled up issues, I didn't enable `ETESYNC_TESTS`.

These test should be rewritten to use docker-based service, or at least in it's own `virtualenv` so we don't have to worry about conflicting dependencies.

WhyNotHugo:
The current [experimental] implementation is stale, and hasn't been
maintained for a long time. Regrettably, not even its tests still work,
and there nobody interested in maintaining it.

If anyone is interested in re-implementing this in a third-party
package, I can consider adding support for pluggable storages.

Closes #875
Closes #827
Closes #638


tasn:
💔 😭

PR #953, opened 2021-12-09
Add shell strategy to fetch params

electrickite:
Adds a new strategy, `shell`, to config fetch parameters. The syntax is nearly identical to the `command` strategy, except that the arguments are run using the `shell=True` subprocess option.

The rationale for this addition is that the output of some password managers (including `pass`) may require piping through additional commands. For example if a `pass` password file is structured as suggested in the documentation:

```
my-secret-password
URL: www.example.com
Username: myuser
```

Then to retrieve only the password, we need to filter the output, for example: `pass my-password | head -n1`

This could be handled on the user side by creating wrapper scripts, but those complicate the use of `vdirsyncer`. Many other tools commonly used in this space (isync, mutt, msmtp, etc.) support password command in shell syntax, so this would remove the need to create custom config scripts for `vdirsyncer`.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -6,6 +6,7 @@ In alphabetical order:
 - Ben Boeckel
 - Christian Geier
 - Clément Mondon
+- Corey Hinshaw
 - Hugo Osvaldo Barrera
 - Julian Mehne
 - Malte Kiefer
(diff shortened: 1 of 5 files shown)


WhyNotHugo (approved):
Looks great, thanks!

WhyNotHugo:
I've re-triggered CI, I don't think failure is related to your changes.

WhyNotHugo:
Thanks, sorry for the delay.

PR #961, opened 2022-01-14
vdirsyncer/metasync.py: for read_only storages adjust the metadata resolution in favour of the read_only storage

dilyanpalauzov:


dilyanpalauzov:
Otherwise the last change calls set_meta on the read-only storage, and fails.

dilyanpalauzov:
I cannot say why this change works for me, but it does exactly what I want.  I want to let everybody modify a calendar, and periodically push upstream the initial content of the calendar, so that all changes made in the meantime get lost.

vdirsyncer/metasync.py
```diff
@@ -55,7 +55,7 @@ async def _resolve_conflict():
         logger.debug(f"B: {b}")
         logger.debug(f"S: {s}")
 
-        if a != s and b != s:
+        if a != s and b != s or storage_a.read_only or storage_b.read_only:
```
WhyNotHugo, line 58: If either collection is read only, then `_resolve_conflict` will get called, but it doesn't do any special handling for read only collections.

I think this works for you because you likely have a `a wins` (or `b wins`) rule.

I think that, without an `_ wins` this would fail, which sounds sensible in this particular scenario.

dilyanpalauzov: I have "b wins" rule.

WhyNotHugo:
Thanks!

Issue #977, opened 2022-04-12
Formula `vdirsyncer` has not implemented #plist, #service or installed a locatable service file


vaboro:
Whenever I run

`brew services start vdirsyncer`

I get

`Error: Formula 'vdirsyncer' has not implemented #plist, #service or installed a locatable service file`

I have Homebrew 3.4.6-46-g1137386 and install vdirsyncer, version 0.18.0

<details>
  <summary>This is what's in the Cellar after install</summary>

**server:~ ladmin$ ls -la /usr/local/Cellar/vdirsyncer/0.18.0/
total 88
drwxr-xr-x  11 ladmin  admin    374 12 apr 18:35 .
drwxr-xr-x   3 ladmin  admin    102 12 apr 18:34 ..
drwxr-xr-x   3 ladmin  admin    102 12 apr 18:35 .brew
-rw-r--r--   1 ladmin  staff    490 11 apr 19:02 AUTHORS.rst
-rw-r--r--   1 ladmin  staff  20715 12 apr 18:34 CHANGELOG.rst
-rw-r--r--   1 ladmin  admin   1759 12 apr 18:35 INSTALL_RECEIPT.json
-rw-r--r--   1 ladmin  staff   1594 11 apr 19:02 LICENSE
-rw-r--r--   1 ladmin  staff   2213 12 apr 18:34 README.rst
drwxr-xr-x   3 ladmin  admin    102 12 apr 18:35 bin
drwxr-xr-x   6 ladmin  admin    204 12 apr 18:34 libexec
-rw-r--r--   1 ladmin  staff   1625 12 apr 18:35 vdirsyncer.plist**
</details>

Everything worked fine with vdirsyncer, version 16.8.0 with config file located at ~/.config/vdirsyncer/config
I tried to checkout vdirsyncer, version 16.8.0 from homebrew-core but then during installation I get:

`Error: vdirsyncer: undefined method 'cellar' for #<BottleSpecification:0x00007f868b143318>`

Looking for ways to resolve the issue with vdirsyncer, version 0.18.0.

WhyNotHugo:
I suggest also opening an issue on the homebrew side.

I do see a `plist` file, so I've no idea what this error means.

vaboro:
I've [opened](https://github.com/Homebrew/discussions/discussions/3176#discussion-4005391) an issue on the homebrew side and I think a solution may be as follows:

[...]

PR #978, opened 2022-04-22
Allow Homebrew to set the value of Label attribute in contrib/vdirsyncer.plist

vaboro:
PR following issue #977

### DIFF
--- a/contrib/vdirsyncer.plist
+++ b/contrib/vdirsyncer.plist
@@ -13,7 +13,7 @@
       <string>@@LOCALE@@</string>
     </dict>
     <key>Label</key>
-    <string>vdirsyncer</string>
+    <string>@@LABEL@@</string>
     <key>WorkingDirectory</key>
     <!-- working directory for vdirsyncer, usually the base directory where
          vdirsyncer is installed, e.g. /usr/local/ -->


WhyNotHugo:
@branchvincent: Since you seem to maintain the homebrew formula, do you mind giving some input on this?

I'm not familiar enough with brew and plists to understand if this is 100% correct, and would prefer some confirmation before merging this in.

branchvincent:
@WhyNotHugo sorry for overlooking this! I've just opened and merged https://github.com/Homebrew/homebrew-core/pull/103983 to fix this on the Homebrew side. In doing so, we no longer use this plist file but rather define it ourselves as we do with other services (which allows us to support both macOS and Linux)

@vaboro when you get a chance, it would be great to confirm this works for you!

vaboro:
@branchvincent thank you for the https://github.com/Homebrew/homebrew-core/pull/103983 merge.
After upgrading vdirsyncer on my machine to 0.18.0_1 the brew vdirsyncer service successfully restarted. I admit that since I'm on macOS I completely overlooked homebrew vdirsyncer's formula cross compatibility with both macOS and Linux; therefore, it seems all the overkill with the revised homebrew formula is really worth it. It also seems there is no need in vdirsyncer.plist in the repo anymore.

PR #985, opened 2022-05-24
gcal: replace oob OAuth2 with local server redirect

### LINKED ISSUE #975: doesn't comply with Google's OAuth 2.0 policy for keeping apps secure
I followed the instructions at https://vdirsyncer.pimutils.org/en/stable/config.html#google to configure synchronization between my Google Contacts and my local filesystem.

```
$ vdirsyncer --version
vdirsyncer, version 0.18.0

$ uname -mprsv
Darwin 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64 i386

$ cat ~/.config/vdirsyncer/config
[general]
status_path = "~/.vdirsyncer/status/"

[pair matt_contacts]
a = "matt_contacts_local"
b = "matt_contacts_remote"

collections = ["from a", "from b"]

metadata = ["displayname"]

[storage matt_contacts_local]
type = "filesystem"
path = "~/.contacts/"
fileext = ".vcf"

[storage matt_contacts_remote]
type = "google_contacts"
token_file = "~/.matt_contacts_remote.vdirsyncer.token_file"
client_id = "..."
client_secret = "..."

$ vdirsyncer discover
```

image

I found the following information indicating that the flow vdirsync is using is deprecated.

> [OAuth out-of-band (oob) flow will be deprecated](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html#disallowed-oob)
OAuth [out-of-band](https://developers.google.com/identity/protocols/oauth2/native-app#manual-copypaste) (OOB) is a legacy flow developed to support native clients which do not have a redirect URI like web apps to accept the credentials after a user approves an OAuth consent request. The OOB flow poses a remote phishing risk and clients must migrate to an alternative method to protect against this vulnerability. New clients will be unable to use this flow starting on Feb 28, 2022.

> [What do I need to do](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html#instructions-oob)
Determine if your app is using the OOB flow
You can inspect your app code or the outgoing network call (in case your app is using an OAuth library) to determine if the Google OAuth [authorization request](https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server) your app is making has the following values for “redirect_uri” parameter.

> redirect_uri=urn:ietf:wg:oauth:2.0:oob or urn:ietf:wg:oauth:2.0:oob:auto or oob

I have not been able to reproduce the success noted in the similar issue https://github.com/pimutils/vdirsyncer/issues/970#issuecomment-1066010007 .

telotortium:
Google Calendar has disabled the oob method for new credentials (see
https://developers.google.com/identity/protocols/oauth2/native-app), so
new users cannot currently use Google Calendar. Fix this by switching to
a loopback redirect_uri flow instead.

This is completely secure - Javascript on the OAuth2 authentication page
opened by the workflow makes a single GET request to a server listening
only on 127.0.0.1. Once the authentication page makes the GET request,
the local web server is not needed anymore and is immediately shut down.

Fix #975

### DIFF
--- a/vdirsyncer/storage/google.py
+++ b/vdirsyncer/storage/google.py
@@ -2,7 +2,10 @@ import json
 import logging
 import os
 import urllib.parse as urlparse
+import wsgiref.simple_server
+import wsgiref.util
 from pathlib import Path
+from threading import Thread
 
 import aiohttp
 import click
@@ -14,6 +17,8 @@ from ..utils import expand_path
 from ..utils import open_graphical_browser
 from . import base
 from . import dav
+from .google_helpers import _RedirectWSGIApp
+from .google_helpers import _WSGIRequestHandler
 
 logger = logging.getLogger(__name__)
 
@@ -54,6 +59,7 @@ class GoogleSession(dav.DAVSession):
         self._client_id = client_id
         self._client_secret = client_secret
         self._token = None
+        self._redirect_uri = None
 
     async def request(self, method, path, **kwargs):
         if not self._token:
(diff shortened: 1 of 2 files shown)


WhyNotHugo (changes requested):
Looks good.

This won't work in some specific scenarios (notably, when `vdirsyncer` is not running on the local host), but I don't think Google has presented any replacement for the API they've deprecated, so it doesn't seem like we have a choice here.

I mostly have minor feedback, I'll try this properly when I have the time, but generally makes sense.

WhyNotHugo:
I've moved all code taken from `google-auth-library-python-oauthlib` into a separate file with a licence header to avoid any conflicts.

I tested this on my local setup and it seems to work, will merge as soon as CI is done running.

WhyNotHugo:
Thanks!

Issue #998, opened 2022-07-29
Restart=on-failure in systemd service is unnecessary


rEnr3n:
OS: Arch Linux
```
$ vdirsyncer --version
vdirsyncer, version 0.18.1.dev0+g3191886.d20211204
```

Wouldn't it make sense to remove `Restart=on-failure` on vdirsyncer.service? It's going to be restarted by the timer anyway. Having `Restart=on-failure` would keep resyncing when one of the remotes is unreachable causing a bit of a grind on disk.

WhyNotHugo:
I guess just retrying once would be good enough?

rEnr3n:
That would be fine too.

WhyNotHugo:
PRs welcome.

rEnr3n:
Do I need to write tests for this? I don't know how to write one.

WhyNotHugo:
No, we don't ship tests for systemd unit files.

PR #999, opened 2022-08-05
Make systemd service restart only once on failure

rEnr3n:
Fixes https://github.com/pimutils/vdirsyncer/issues/998


WhyNotHugo:
Thanks

Issue #1022, opened 2022-12-27
No module named 'aiostream' in python39+


lduchosal:
Before you submit bug reports: https://vdirsyncer.pimutils.org/en/stable/contributing.html

* vdirsyncer, version 0.19.0 
* freebsd port py39-vdirsyncer
* python Python 3.9.16
* FreeBSD 12.4

error: Unknown error occurred: No module named 'aiostream'
error: Use `-vdebug` to see the full traceback.
debug:   File "/usr/local/lib/python3.9/site-packages/vdirsyncer/cli/__init__.py", line 32, in inner
debug:     f(*a, **kw)
debug:   File "/usr/local/lib/python3.9/site-packages/vdirsyncer/cli/__init__.py", line 126, in sync
debug:     from .tasks import prepare_pair
debug:   File "/usr/local/lib/python3.9/site-packages/vdirsyncer/cli/tasks.py", line 8, in <module>
debug:     from .discover import collections_for_pair
debug:   File "/usr/local/lib/python3.9/site-packages/vdirsyncer/cli/discover.py", line 8, in <module>
debug:     import aiostream

Replace :

import aiostream

With :
if sys.version_info < (3,9):
   import aiostream

Fixes the bug, Would you accept PR?

WhyNotHugo:
You need to install `aiostream`. Are you installing via `pip`? If so, use `pip install aiostream`.

PR #1023, opened 2022-12-27
Fix: import aiostream only if sys.version < python 3.9

lduchosal:


WhyNotHugo:
If you don't import the module, then the code will fail to run since this module is used in quite a few places.

Why would you skip it for python < 3.9? What issue are you trying to solve here?

lduchosal:
issue #1022 sum up my problem. aiostream is not needed in python 39 .. Everything works fine if imports are removed.

WhyNotHugo:
I don't know why you assume that it's not necessary, but it is. There's a lot of code relying on it (just look for other references to `aiostream`).

**Most** of the usages of this library is in tests, but not all of them.

Issue #184, opened 2015-02-25
Exclude collections from syncronization


shervin86:
The special values "from a" and "from b" in the "collections" field is very useful. 
It would be nice to have a way to exclude specific collections, like:
```
[pair zimbra_calendar]
a = zimbra_calendar_local
b = zimbra_caldav_remote
collections = ["from b" "!work" ]
```
where all the collections in b are downloaded except the one that is called "work"

untitaker:
I don't think I want this functionality, at least as long as it's not requested multiple times.

The collections parameter originally didn't have "from a" or "from b". These parameters were only added for the particular purpose of syncing all collections. It was also intended for cases where people didn't know the collection names. I don't really want to add complex black/whitelisting or set operations to the config syntax.

Do this instead:

  1.  Run `vdirsyncer discover`. It will discover all collections, and give you all their names.
  2. Paste the collection names into your config.

c-schwamborn:
I could also use this feature as I'm undertaking a huge migration at my University where i need to expunge some unwanted calendars present in every principal. The suggestion above with `vdirsyncer discover` is quite unpractical in this case.

PR #1025, opened 2022-12-31
add some blacklist code to allow exclude collections from syncronization #184

c-schwamborn:
I hope this isn't to crude as I didn't had much time to dive to deep into the code.

### DIFF
--- a/vdirsyncer/cli/config.py
+++ b/vdirsyncer/cli/config.py
@@ -60,6 +60,7 @@ def _validate_collections_param(collections):
         raise ValueError("`collections` parameter must be a list or `null`.")
 
     collection_names = set()
+    collection_blacklist = set()
 
     for i, collection in enumerate(collections):
         try:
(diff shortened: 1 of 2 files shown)


vdirsyncer/cli/config.py
```diff
@@ -87,9 +88,14 @@ def _validate_collections_param(collections):
             if collection_name in collection_names:
                 raise ValueError("Duplicate value.")
             collection_names.add(collection_name)
```
WhyNotHugo, line 90: If a name starts with `!`, you're also adding it to the `collection_names` list here. You probably want to move this line to an `else` block after `if collection[0] == "!":`.

WhyNotHugo (commented):
General idea looks good. Please name this the parameter / list `excludes` (or maybe `ignored`?). I'd prefer to avoid the term `blacklist`.

vdirsyncer/cli/config.py
c-schwamborn: Yes I saw that, but does it matter at this point, as `collection_names` is temporary here and as I understand only used to weed out duplicate values from the collections list. When I leave the `excludes` within the `collection_names` I don't have to bother further with that. Or did I missed something there and `collection_names` is used for anything else?

I replaced `blacklist` with `excludes`, comes in the next push

PR #1029, opened 2023-01-06
Corrected installation steps for Ubuntu and pip

chrysle:
I corrected the python source package for Ubuntu to be `python3`, and removed the pip `--user` option, since this is the default now: 

```
# After installing vdirsyncer without --user option 
$ pip list -v | grep "vdirsyncer"
vdirsyncer              0.19.0               /home/USER/.local/lib/python3.8/site-packages pip 
```


WhyNotHugo:
Looks good, thanks!

PR #1031, opened 2023-01-13
Enable environment variables for HTTP proxy

daniele-athome:
I don't know why it wasn't enabled, but here it is.


WhyNotHugo (approved):
Looks good, thanks!

PR #1038, opened 2023-01-30
Use ruff for code checking and linting

WhyNotHugo:
`isort` is failing in pre-commit.ci right now, so this might be good timing.

See: `isort` is failing in pre-commit.ci, so this might be good timing.


WhyNotHugo:
Aw shit, a test is broken, and totally unrelated. Needs bisecting.

Issue #814, opened 2020-04-02
Sync Google Contacts with Nextcloud - No Groups


etofi:
vdirsyncer version: 0.16.7
Python version: 3.8.2

I synchronize my Google contacts with my NextCloud. Works well.
The only problem is that the groups are not synchronized.

WhyNotHugo:
Key issue is that groups are not fetched.

Not sure if the Google API exposes these, if it does, we should fetch them, otherwise, document the caveat.

timoludwig:
This seems to be the related issue at Google: https://issuetracker.google.com/issues/36761530

Since there have been no real updates for 6 years now, I assume this is labeled as "wontfix" internally or has a veeeery low priority for them.

One workaround would be to use the people API to retrieve the contact groups and merge this information into the vCard files, but not sure whether it's worth the effort (and also this makes syncing much slower)...

axet:
Recent Google Contacts showing no Groups, but Labels instead. Also no CATEGORIES in .vcf files created by vdirsyncer

WhyNotHugo:
Based on that last comment, is there anything to be done for this issue, or are groups properly exported as categories.

PR #1050, opened 2023-03-05
Document caveats of Google contacts storage

timoludwig:
Even though it's unclear whether this might be fixed in the future, I guess a note in the docs would be beneficial as long as it isn't.

Thanks in advance for your feedback!

Reference: #814

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -268,6 +268,14 @@ in terms of data safety**. See `this blog post
 <https://evertpot.com/google-carddav-issues/>`_ for the details.  Always back
 up your data.
 
+Another caveat is that Google group labels are not synced with vCard's
+`CATEGORIES <https://www.rfc-editor.org/rfc/rfc6350#section-6.7.1>`_ property
+(also see :gh:`814` and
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Can you also include a link to the upstream issue please? That's https://issuetracker.google.com/issues/36761530

timoludwig:
Alright, done!
I just couldn't find an upstream issue for the BDAY problem, just a few other projects struggling with the same problem:
- https://gitlab.gnome.org/GNOME/evolution/-/issues/1287#note_1100346
- https://gitlab.com/CardBook/CardBook/-/issues/866

WhyNotHugo:
Thanks!

WhyNotHugo:
>  I just couldn't find an upstream issue for the BDAY problem, just a few other projects struggling with the same problem:

I'd suggest reporting this upstream, but I don't think Google cares about these bugs, the other bug is 6 years old. And it's the kind of bug where apparently data is simply lost. :shrug:

timoludwig:
Alright, I opened https://issuetracker.google.com/issues/271731853 for this, if I get a response I will update you.

Issue #818, opened 2020-05-01
vdirsyncer single file calendar sync's growth of running time is quadratic or higher (it takes "forever")


bernhardreiter:
Observed:
running ` /usr/bin/vdirsyncer -v debug sync` and the process hangs in one thread.

Expected: Sync completes. It did before the move to Debian buster and switch of vdirsyncer version. Or a diagnostic message that help finding the problem, if there is one.

Further observations:
an strace shows `futex(0x7f387429e890, FUTEX_WAIT_BITSET_PRIVATE|FUTEX_CLOCK_REALTIME, 0, NULL, FUTEX_BITSET_MATCH_ANY` and a py-bt with gdb shows 
```
  File "/usr/lib/python3.7/threading.py", line 296, in wait
    waiter.acquire()
  File "/usr/lib/python3.7/queue.py", line 170, in get
    self.not_empty.wait()
  File "/usr/lib/python3/dist-packages/click_threading/__init__.py", line 110, in run
    func, future = self.tasks.get()
  File "/usr/lib/python3/dist-packages/vdirsyncer/cli/utils.py", line 375, in join
    ui_worker.run()
  <built-in method next of module object at remote 0x7f387d2bcc28>
  File "/usr/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/usr/lib/python3/dist-packages/vdirsyncer/cli/__init__.py", line 154, in sync
    wq.spawn_worker()
[..]
```

* vdirsyncer 0.16.7 (problem reproducable with pipx install from pypi and Debian Buster version)
* posteo server (but probably irrelevant)
* python3 Debian Buster Version: 3.7.3-1
* config:
```
[general]
status_path = "~/.vdirsyncer/status/"

[pair my_calendar]
a = "my_calendar_remote"
b = "my_calendar_local"
collections = null
conflict_resolution = "a wins"

[storage my_calendar_remote]
type = "caldav"
read_only = true
url = "https://posteo.de:8443/calendars/bernh/default/"
username = XXX@posteo.de"
password = "YYY"

[storage my_calendar_local]
type = "singlefile"
path = "~/.vdirsyncer/my_calendar.ics"
```
* the last line of -v debug is
```
debug: Already normalized: '/calendars/bernh/default/XXXXX.ics'
```

* Related reports: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959097 with more details.

bernhardreiter:
[...]

PR #1051, opened 2023-03-10
Prevent single file storage from performing unnecessary N^2 loop

eguiraud:
For single file storage we wrap the logic of get_multi with the at_once context manager so that `self.list()` (which is called by `self.get()`) actually caches the items rather than re-computing them at every call.

This should largely mitigate the performance issue describe at https://github.com/pimutils/vdirsyncer/issues/818 . The issue discussion also contains more background about this patch.

### DIFF
--- a/vdirsyncer/storage/singlefile.py
+++ b/vdirsyncer/storage/singlefile.py
@@ -4,6 +4,7 @@ import functools
 import glob
 import logging
 import os
+from typing import Iterable
 
 from atomicwrites import atomic_write
 
@@ -11,6 +12,7 @@ from .. import exceptions
 from ..utils import checkfile
 from ..utils import expand_path
 from ..utils import get_etag_from_file
+from ..utils import uniq
 from ..vobject import Item
 from ..vobject import join_collection
 from ..vobject import split_collection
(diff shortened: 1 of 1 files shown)


WhyNotHugo (approved):
Perfect, thanks!

WhyNotHugo:
CI is failing because the cache is _never_ initialized during `multi-get`, even if it's not already.

It always needs to happen if `self._items is None`.

WhyNotHugo:
I wonder if the failure scenario is only possible in tests or also happens in some untested real-life scenario.

eguiraud:
I don't understand how this can happen:

```
vdirsyncer/storage/singlefile.py:138: in get_multi
    async with self.at_once():
/usr/lib/python3.10/contextlib.py:206: in __aexit__
    await anext(self.gen)
vdirsyncer/storage/singlefile.py:200: in at_once
    self._write()
vdirsyncer/storage/singlefile.py:186: in _write
    text = join_collection(item.raw for item, etag in self._items.values())
E   AttributeError: 'NoneType' object has no attribute 'values'
```

`at_once` calls `self.list()` first thing, and `list()` calls `self._items = collections.OrderedDict()`. So I would expect `_items` to be initialized to an empty dict by the time `at_once` calls `_write`?

WhyNotHugo:
See b1ef68089b11eed1050542108296b9a8d7dd0ae6

PR #1061, opened 2023-04-11
Add ldap storage

nblock:
The ldap backend is currently limited to a few attributes (mostly phone related) but I'm willing to expand this to other vcard fields. It is currently being tested with an OpenLDAP server and supports the `inetOrgPerson` schema. In my test setup I'm syncing from Nextcloud to OpenLDAP and from OpenLDAP to filesystem. This PR is not complete yet, but I'm looking for feedback.

Tasks:

* [x] List entries in LDAP
* [x] Get an entry from LDAP
* [x] Create new entry in LDAP
* [x] Update an existing entry in LDAP
* [x] Delete an entry from LDAP
* [ ] Support metadata?
* [ ] Support collections?
* [ ] Fix CI builds
* [ ] Fix tests
* [ ] Documentation
* [ ] Cleanup, squash and proper commit message

This is more or less a continuation of #500 by @paulfariello (thanks for your effort!).

See also: #500 
Fixes: #283

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
     # Optional dependencies
     extras_require={
         "google": ["aiohttp-oauthlib"],
+        "ldap": ["ldap3", "vobject"],
     },
     # Build dependencies
     setup_requires=["setuptools_scm != 1.12.0"],
--- a/tests/storage/test_ldap.py
+++ b/tests/storage/test_ldap.py
@@ -1,0 +1,32 @@
+import ldap3
+import pytest
+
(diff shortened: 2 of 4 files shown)


WhyNotHugo:
I'm not very familiar with LDAP, but don't see anything obviously wrong here.

I'm curious about the LDAP<>vcard conversion that you're doing. Some fields in vcard don't seem to be supported in LDAP. This means that when you sync vcard->ldap the fields are dropped, but if things later change on the LDAP side and the contact is synced ldap->vcard those dropped fields will be lost forever, right?

PR #1076, opened 2023-06-26
ci: use virtualenvs for jobs that use pip

WhyNotHugo:
Pip now refuses to tamper with the system python installation.


WhyNotHugo:
This will fix `tests-pypi`, but no `tests-minimal`. See: https://github.com/pimutils/vdirsyncer/issues/1077

PR #1078, opened 2023-07-07
Update config.rst

frosklis:
Setting up calendar synchronization for the first time in google did not work following the docs.This is the way it actually works:
With web application, a 400 is returned. With desktop it works fine


WhyNotHugo:
Indeed, this is correct. "Web application" fails with a 400 error, and "Desktop application" seems to work fine.

PR #1080, opened 2023-07-14
Add typing annotations

just1602:
I was browsing the commit history since I notice there was a new release, and I've seen there was a commit adding some type annotation, so I try to add some typing myself.

I mainly added types in the `cli/config.py` file and I run `ruff` which fix a couple of things.

I hope you find the PR useful, otherwise I won't be offended if you close it.


WhyNotHugo:
Thanks! These help understanding the code a lot so are always welcome (as long as they make sense).

just1602:
Good!

It's not always easy to type stuff since some variables can be more then one thing, and this is why the type I had where not perfect, but if I get time, I'll try to add more of them.

PR #1081, opened 2023-07-30
fix(repair_collection): use DiscoverResult logic to discover collection (#1019)

Croissong:
Fixes error: `DAVSession.__init__() missing 1 required keyword-only argument: 'connector'` (#1019).
Reuses the existing logic in `DiscoverResult` to determine if the storage requires a `connector` arg.

Please let me know if there are any tests that I can extend without too much effort :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,11 @@ Package maintainers and users who have to manually update their installation
 may want to subscribe to `GitHub's tag feed
 <https://github.com/pimutils/vdirsyncer/tags.atom>`_.
 
+Version 0.19.3
+==============
+
+- Fix crash when running ``vdirsyncer repair`` on a collection. :gh:`1019`
+
 Version 0.19.2
 ==============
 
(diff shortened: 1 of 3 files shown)


WhyNotHugo (approved):
I tested this on a caldav collection of mine and it fixes the issue as promised. :+1:

WhyNotHugo:
Do you want to add an entry to `CHANGELOG.rst` for 0.19.3? Feel free to add yourself to `AUTHORS.rst` as well.

Croissong:
done :heavy_check_mark:
Not sure if this bug affected only DAV or also other storage backends.

WhyNotHugo:
I think it affected only dav; I could not reproduce the issue with a `filesystem` storage.

WhyNotHugo:
Thanks!

zetoph:
Hello, thanks a lot for this fix, it works perfectly! Do you already know when this might be published to the pip or apt repos? When installing using those, I still get the old version.
br

PR #1085, opened 2023-08-23
Add type hint to vdirsyncer/cli/utils.py

just1602:
Hello,

I added some type hint in the `vdirsyncer/cli/utils.py` file. This also allowed me to remove some `Ǹone` check in the code, but I've made those modifications in separate commits, so if you only want the type hinting without the other refactors, you can.


vdirsyncer/cli/utils.py
```diff
@@ -174,10 +181,15 @@ def get_status_path(base_path, pair, collection=None, data_type=None):
     return path
 
 
-def load_status(base_path, pair, collection=None, data_type=None):
+def load_status(
+    base_path: str,
+    pair: str,
+    collection: str | None = None,
+    data_type: str | None = None,
+) -> dict[str, Any] | None:
```
WhyNotHugo, line 189: With the changes made below, this function can no longer return `None`.

vdirsyncer/cli/tasks.py
```diff
@@ -142,11 +142,8 @@ async def metasync_collection(collection, general, *, connector: aiohttp.TCPConn
     try:
         cli_logger.info(f"Metasyncing {status_name}")
 
-        status = (
-            load_status(
-                general["status_path"], pair.name, collection.name, data_type="metadata"
-            )
-            or {}
+        status = load_status(
+            general["status_path"], pair.name, collection.name, data_type="metadata"
```
WhyNotHugo, line 146: **Nitpick**: can you include a trailing `,` here so that these arguments get auto-formatted into something easier to read?

vdirsyncer/cli/utils.py
```diff
@@ -4,6 +4,7 @@
 import json
 import os
 import sys
+from typing import Any
```
WhyNotHugo, line 7: I think we also need `from __future__ import annotations` in this file for compat with older Pythons.

This likely applies to other files touched in the PR too.

WhyNotHugo (commented):
Just minor changes required. Overall this looks good, and keeping it all in the same PR is fine by me.

vdirsyncer/cli/utils.py
WhyNotHugo: CI failures are because of this: https://builds.sr.ht/~whynothugo/job/1045797#task-test

WhyNotHugo:
Mypy is now warnings of some type violations too: https://results.pre-commit.ci/run/github/16865424/1692764316.Jvj2-1qBQQ6LKIiGJNyXWg

vdirsyncer/cli/utils.py
just1602: I think it's a mistake I made while rebasing some ruff fixes. I.ll fix all that.

WhyNotHugo (approved):
Lovely, thanks!

PR #1090, opened 2023-09-27
Add a no_delete option to the storage configuration

### LINKED ISSUE #1088: Feature request: add a no_delete option to the storage configuration
I would like to add a no_delete option to the storage configuration.

The read_only option did not work for our use case:
1) I want to copy the calendar entries/.ics files that are in multiple calendars A, B, C etc to one calendar X. So, in calendar A are a few one calendar entries, and copying them to calendar X with vdirsyncer is no problem.
2) Then I want to copy the calendar entries of calendar B to calendar X. This works, but the calendar entries, which were copied from calendar A to calendar X, are deleted from calendar X, because they don't exist in calendar B.
3) If the read_only option which is enabled for calendars A, B, C would be enabled for calendar X as well, it wouldn't be possible to copy _any_ calendar entry to X.

I have implemented the function in our university without any problems. Since one should wait for the discussion in the issue tracker first, I don't make a pull request yet.

KaiHerlemann:
Fixes #1088

### DIFF
--- a/vdirsyncer/sync/__init__.py
+++ b/vdirsyncer/sync/__init__.py
@@ -241,7 +241,9 @@ class Delete(Action):
 
     async def _run_impl(self, a, b):
         meta = self.dest.status.get_new(self.ident)
-        if not self.dest.storage.read_only and not self.dest.storage.no_delete:
+        if self.dest.storage.read_only or self.dest.storage.no_delete:
+            sync_logger.debug(f"Skipping deletion of item {self.ident} from {self.dest.storage}")
+        else:
             sync_logger.info(f"Deleting item {self.ident} from {self.dest.storage}")
             await self.dest.storage.delete(meta.href, meta.etag)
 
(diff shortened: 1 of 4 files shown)


vdirsyncer/sync/__init__.py
```diff
@@ -241,7 +241,7 @@ def __init__(self, ident, dest):
 
     async def _run_impl(self, a, b):
         meta = self.dest.status.get_new(self.ident)
-        if not self.dest.storage.read_only:
+        if not self.dest.storage.read_only and not self.dest.storage.no_delete:
```
WhyNotHugo, line 244: I'd rather log this just to be clear. Also simplifies the `if` condition.

```suggestion
        if self.dest.storage.read_only or self.dest.storage.no_delete:
            sync_logger.debug(f"Skipping deletion of item {self.ident} from {self.dest.storage}")
        else:
```

WhyNotHugo (changes requested):
LGTM, I'd only make a minor change here.

vdirsyncer/sync/__init__.py
KaiHerlemann: works fine for me.

WhyNotHugo:
LGTM

WhyNotHugo:
Thanks. Merged via ebaeeb80f255ac3528c429ef3f906118c634b0f7, with a minor tweak to fix a failing test.

PR #1097, opened 2023-11-24
docs: add instructions to get pw from environment variable

azrdev:
tested with vdirsyncer 0.19.2 on archlinux


WhyNotHugo:
I had no idea about the `printenv` command, thanks for the hint.

PR #1100, opened 2023-12-09
Update config.rst

meeuw:
It's required to set collections = null now


meeuw:
Oh sorry for the late reply, the context is above:
> Example for syncing with :storage:`caldav` using a ``null`` collection::

It's now required to set collection to null if you want to use a null collection. It doesn't seem to be a default anymore.

PR #1106, opened 2023-12-20
[WIP] Add options remove_details and required_attendee

DavidMikeSimon:
I added these features for my own self-hosted setup, and was wondering if they're something you would want to merge upstream. If so, I can write some unit tests and clean this up a bit.


DavidMikeSimon:
@WhyNotHugo Thoughts on whether these features would be useful to upstream?

WhyNotHugo:
I'm not sure that this is something that I'd want inside vdirsyncer. Generally the focus is to synchronise items without any changes. I fear that merging `remove_details` would lead to other requests to exclude other individual fields and lead to too much feature creep.

To understand your use-case better: do you want two-way sync, or do you just want work items to be synchronised to another calendar one-way?

DavidMikeSimon:
@WhyNotHugo My use case is all about one-way sync. I work from home, and my workplace uses Google Calendar. I also have a home calendar system using NextCloud, and I want my family to be able to use that to see what times during each day I will be busy in a meeting. The `required_attendee` feature is necessary for this, because my work calendar has meetings that I don't attend, but I cannot delete those events because they are shared.

The `remove_details` feature isn't necessary for functionality, but I like it for safety. I want to avoid the risk of exposing sensitive data from my work if someone were to break into my instance of NextCloud.

DavidMikeSimon:
What if `remove_details` accepted a list of fields to remove, rather than having a hard-coded set?

PR #1116, opened 2024-02-01
A bit more explanation of "from a" and "from b"

zeroheure:


WhyNotHugo:
Thanks

WhyNotHugo:
Sorry, I didn't expect that pressing <kbd>ctrl</kbd>+<kbd>enter</kbd> would "close and comment" instead of just "comment".

PR #1117, opened 2024-02-01
nested pairs can be used

zeroheure:


WhyNotHugo:
Can you clarify what you mean here?

PR #1120, opened 2024-02-26
[WIP] Handle recurring events with duplicate UIDs

DavidMikeSimon:


WhyNotHugo:
What  is the issue that you're addressing?

DavidMikeSimon:
@WhyNotHugo Still working on this, but basically the problem is that `UID` isn't guaranteed to be unique. See https://stackoverflow.com/questions/62982636/in-what-situations-could-an-ical-vevent-end-up-with-a-duplicated-uid

DavidMikeSimon:
See also this text from RFC 5546:

>    1.  The primary key for referencing a particular iCalendar component
       is the "UID" property value.  To reference an instance of a
       recurring component, the primary key is composed of the "UID" and
       the "RECURRENCE-ID" properties.

WhyNotHugo:
If multiple events are instances of the same recurring event, they share the same UID.

From https://www.rfc-editor.org/rfc/rfc4791#section-4.1:

>   The UID property value of the calendar components contained in a
>   calendar object resource MUST be unique in the scope of the calendar
>   collection in which they are stored.

This implies that is multiple instances of a single event exist, they MUST be stored in the same calendar object resource.

Also:

WhyNotHugo:
E.g.: the would have the same `href`.

DavidMikeSimon:
@WhyNotHugo Yeah, I think you're right! I was looking for the problem in the wrong place.

I'm attempting to synchronize events from an ICS from a Google Calendar URL to a local CalDav server. I noticed that some instances of recurring events were not present on the destination.

To narrow it down, I just now tried to reproduce the issue in a sync from the downloaded ICS file as type `singlefile` to a directory as type `filesystem`. And I noticed it doing exactly as you said: putting the recurring events with the same UID in the same file. So I believe the problem is not in `vdirsyncer` but in my receiving server.

Sorry about the distraction, and thank you for the info! I'll go ahead and close this PR.

PR #1121, opened 2024-03-18
fix pylint W0621: Redefining name 'main' from outer scope (line 68) (redefined-outer-name)

meeuw:
This fixes pylint W0621: Redefining name 'main' from outer scope (line 68) (redefined-outer-name)


WhyNotHugo:
Sure, makes sense. Thanks.

PR #1122, opened 2024-04-01
docs: update config info on todoman tutorial

octvs:
stumbled upon this while setting up todoman


WhyNotHugo:
Nice catch, thanks.

Issue #1021, opened 2022-12-23
password.fetch command applies path normalization to non-path parameters


s-hamann:
## Problem Description

I configured vdirsyncer to get the password from the user keyring:
```
password.fetch = ["command", "secret-tool", "lookup", "URL", "https://example.com/"]
```

This used to work fine before version 0.19.0. Since 0.19.0, however, the configured parameters get changed and, thus, `secret-tool` does not find the password any more:
```
vdirsyncer -vdebug sync
debug: Fetching value for password.fetch with command strategy.
error: Unknown error occurred: Command '['secret-tool', 'lookup', 'URL', 'https:/example.com']' returned non-zero exit status 1.
error: Use `-vdebug` to see the full traceback.
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/__init__.py", line 32, in inner
debug:     f(*a, **kw)
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/__init__.py", line 150, in sync
debug:     asyncio.run(main(collections))
debug:   File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
debug:     return loop.run_until_complete(main)
debug:   File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
debug:     return future.result()
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/__init__.py", line 133, in main
debug:     async for collection, config in prepare_pair(
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/tasks.py", line 24, in prepare_pair
debug:     await collections_for_pair(
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/discover.py", line 57, in collections_for_pair
debug:     cache_key = _get_collections_cache_key(pair)
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/discover.py", line 32, in _get_collections_cache_key
debug:     pair.config_b,
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/utils.py", line 158, in __get__
debug:     obj.__dict__[self.__name__] = result = self.fget(obj)
debug:   File "/usr/lib/python3.10/site-packages/vdirsyncer/cli/config.py", line 271, in config_b
debug:     return self._config.get_storage_args(self.name_b)
[...]

PR #1124, opened 2024-04-15
don't normalize paths in expand_path

Witcher01:
Remove normalizing paths in "expand_path" and expand environment variables. Paths don't need to be normalized, but doing this on non-path parameters (i.e. URLs) might cause bugs.

Reported-by: s-hamann
Closes: https://github.com/pimutils/vdirsyncer/issues/1021

### DIFF
--- a/vdirsyncer/utils.py
+++ b/vdirsyncer/utils.py
@@ -20,9 +20,9 @@ _missing = object()
 
 
 def expand_path(p: str) -> str:
-    """Expand $HOME in a path and normalise slashes."""
+    """Expand $HOME/~(user) and environment variables in a path."""
     p = os.path.expanduser(p)
-    p = os.path.normpath(p)
+    p = os.path.expandvars(p)
     return p
 
 


WhyNotHugo:
I'd have to recall why we normalised paths in the first place. The initial commit isn't very helpful: da8bba89de3baf26e45f7c70761326277d7532ea But I have a vague impression that it makes a difference in some situations (e.g.: when a user-provided path is compared with an internally-generated path maybe?).

But yeah, we shouldn't normalise paths on URLs, that's outright wrong.

Shadow53:
Looks like the readthedocs job is preventing this PR from being merged? It's saying the `.readthedocs.yaml` file can't be found in the repository, which I don't see on `main` either.

Would like to see this merged as I am also running into this issue.

WhyNotHugo:
I reviewed this and https://github.com/pimutils/vdirsyncer/issues/1021 a bit further.

If a password has `$HOME`, we'd still be messing it up. I think that we need to stop calling `expand_path` on some types of fields.

WhyNotHugo:
Exactly which scenario does this patch fix?

Shadow53:
> Exactly which scenario does this patch fix?

For me, it's the path normalization issue. I use 1Password, and the cli command to get a password is `op read op://<vault>/<item>/password`. The URI bit gets normalized to `op:/<vault>/...`, which causes the command to fail.

I can use a script to hide the real command so this doesn't happen, but I'd rather that not be a long-term solution.

Shadow53:
> If a password has `$HOME`, we'd still be messing it up. I think that we need to stop calling `expand_path` on some types of fields.

Since both `command` and `shell` are supported, one possible option is to let the shell expand environment variables and have `command` pass all arguments verbatim.

PR #1135, opened 2024-09-01
storage/http: expose `ignore_uids` config parameter

pacien:
This allows users to choose whether or not to ignore the UIDs coming
from WebCAL providers and instead use the item's content hash. This
remains the default behaviour.

In my case, my provider was generating stable UIDs but was patching the
export date in other fields, resulting in the hash of all events
changing at each synchronisation, having the opposite effect.


WhyNotHugo:
https://github.com/pimutils/vdirsyncer/pull/1136 doesn't address the issue which you're describing?

PR #1136, opened 2024-09-01
storage/http: add support for `filter_hook`

pacien:
This allows users to process fetched items through a filter command,
to fix malformed webcal items as they are imported.

In my case, my provider adds the export time to the description and
random sequence numbers to all events. This caused the whole collection
to be invalidated and propagated at each sync. I use the filter to
remove those, canonicalising the items.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Version 0.19.3
 - Require matching ``BEGIN`` and ``END`` lines in vobjects. :gh:`1103`
 - A Docker environment for Vdirsyncer has been added `Vdirsyncer DOCKERIZED <https://github.com/Bleala/Vdirsyncer-DOCKERIZED>`_.
 - Implement digest auth. :gh:`1137`
+- Add ``filter_hook`` parameter to :storage:`http`. :gh:`1136`
 
 Version 0.19.2
 ==============
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -484,6 +484,7 @@ leads to an error.
         [storage holidays_remote]
         type = "http"
         url = https://example.com/holidays_from_hicksville.ics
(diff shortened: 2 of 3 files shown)


WhyNotHugo:
Thanks

mdrovdahl:
Great addition. Thanks. 

After some trial and error, I got this working and figured I'd document it.

The command which runs on the filter_hook needs to read stdin, process, and then write to stdout. The output needs to be a valid object of the same type as the input...so, in my case of wanting to retain some events and discard others while syncing, this seems to be working:

```
#!/usr/bin/env python3

import sys
from icalendar import Event, Calendar

def filter_hook(input_data):
    """
    Custom filter_hook implementation for vdirsyncer.
    
    Filters out calendar events that do not contain a specific keyword
    in their summary (title).
    
    Args:
        item (str): The raw iCalendar data as a string.

    Returns:
        str | None: The original item if it passes the filter, otherwise None.
    """

    try:
        # Parse the item using the iCalendar library
        cal = Calendar.from_ical(input_data)

        # Create an empty Calendar & event to return if needed
        new_cal = Calendar()

        # Iterate through the events in the calendar data
        for component in cal.walk('VEVENT'):
            if isinstance(component, Event):
                summary = component.get('SUMMARY', '').lower()
                
                # Define a keyword to filter by
                keyword = "Keyword"
                
                if keyword.lower() in summary:
                    return input_data # Return the input data
        
        # If no events match the criteria, filter out the item
        return new_cal.to_ical().decode("utf-8") # Return an empty, but valid Calendar event
    
    except Exception as e:
        # Log or handle parsing errors
        print(f"Error parsing item: {e}")
        return None

if __name__ == "__main__":
    input_data = sys.stdin.read()
    processed_data = filter_hook(input_data)
    print(processed_data)
```

PR #1137, opened 2024-09-10
Implement digest auth

malmeloo:
This PR makes the following changes:

- Re-implements digest auth by wrapping requests's `HTTPDigestAuth` and using it with aiohttp;
- Removes `requests_toolbelt` since it is no longer used;
- Shows a helpful (critial) error for `guess` auth to indicate that it is not supported;
- Updates the test for geuss auth to reflect its status.

Things to consider:
- The wrapper accesses an internal instance variable of `HTTPDigestAuth`, so it might be a good idea to pin requests to a specific version.

I can personally confirm that both basic and digest auth appear to work with a local baikal server, but more testing may be needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,9 +16,6 @@ requirements = [
     "click>=5.0,<9.0",
     "click-log>=0.3.0, <0.5.0",
     "requests >=2.20.0",
-    # https://github.com/sigmavirus24/requests-toolbelt/pull/28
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
LGTM, thanks

WhyNotHugo:
https://pypi.org/project/vdirsyncer/0.19.3/

malmeloo:
Wonderful, thanks 🙂

Issue #1149, opened 2024-12-05
TestVobjectMachine is failing


lahwaacz:
Hi, when we build the [Arch Linux package](https://gitlab.archlinux.org/archlinux/packaging/packages/vdirsyncer) and run the test suite, one of the tests currently fails:

```
__________________________ TestVobjectMachine.runTest __________________________

self = <hypothesis.stateful.VobjectMachine.TestCase testMethod=runTest>

    def runTest(self):
>       run_state_machine_as_test(cls, settings=self.settings)

/usr/lib/python3.12/site-packages/hypothesis/stateful.py:437:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/lib/python3.12/site-packages/hypothesis/stateful.py:241: in run_state_machine_as_test
    run_state_machine(state_machine_factory)
/usr/lib/python3.12/site-packages/hypothesis/stateful.py:108: in run_state_machine
    @given(st.data())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = VobjectMachine({'parsed': [VarReference(name='parsed_0')],
 'unparsed': [VarReference(name='unparsed_0')]})
c = <vdirsyncer.vobject._Component object at 0x718d86f17fb0>, key = '0'
value = '1'

    @rule(c=Parsed, key=uid_strategy, value=uid_strategy)
    def add_prop(self, c, key, value):
        c[key] = value
>       assert c[key] == value
E       AssertionError: assert '0' == '1'
E
E         - 1
E         + 0
E       Falsifying example:
E       state = VobjectMachine()
E       unparsed_0 = state.get_unparsed_lines(encoded=False, joined=False)
E       parsed_0 = state.parse(unparsed=unparsed_0)
E       state.add_prop_raw(c=parsed_0, key='0', params=[], value='0')
E       state.add_prop_raw(c=parsed_0, key='0', params=[], value='0')
E       state.add_prop(c=parsed_0, key='0', value='1')
E       state.teardown()

tests/unit/utils/test_vobject.py:334: AssertionError
```

There is also a bunch of warnings:

```
=============================== warnings summary ===============================
tests/conftest.py:48
[...]

PR #1153, opened 2024-12-20
Fix _Component.__delitem__ with adjacent identical keys

cjwatson:
Hypothesis found the following example:

```
tests/unit/utils/test_vobject.py:335: in add_prop
    assert c[key] == value
E   AssertionError: assert '0' == '1'
E
E     - 1
E     + 0
E   Falsifying example:
E   state = VobjectMachine()
E   unparsed_0 = state.get_unparsed_lines(encoded=False, joined=False)
E   parsed_0 = state.parse(unparsed=unparsed_0)
E   state.add_prop_raw(c=parsed_0, key='0', params=[], value='0')
E   state.add_prop_raw(c=parsed_0, key='0', params=[], value='0')
E   state.add_prop(c=parsed_0, key='0', value='1')
E   state.teardown()
```

After the two `add_prop_raw` calls, `c.props` is `["0;:0", "0;:0", "FOO:YES"]`.  `_Component.__delitem__` then fails to effectively delete the previous key: it deletes the first `"0;:0"` item, but then checks for continuation lines following it and incorrectly keeps the second `"0;:0"` item even though it begins with one of the prefixes it's trying to delete.  Checking for the prefix in the check for continuation lines fixes this.

Fixes: #1149

### DIFF
--- a/vdirsyncer/vobject.py
+++ b/vdirsyncer/vobject.py
@@ -329,7 +329,7 @@ class _Component:
                 break
 
             for line in lineiter:
-                if not line.startswith((" ", "\t")):
+                if not line.startswith((" ", "\t", *prefix)):
                     new_lines.append(line)
                     break
 


cjwatson:
The codecov failure looks unrelated to my change, as far as I can tell.

WhyNotHugo:
Thanks for figuring this out!

cjwatson:
No problem.  Thanks for including the regression test too!

PR #1162, opened 2025-02-13
atomicwrites: remove dependency on abandoned library

mathstuf:


mathstuf:
Seems like CI has bitrotted some.

mathstuf:
Would it be possible to get a tag with this? I'd like to start pushing the `atomicwrites` Python package from Fedora and a tagged release with this would be very helpful.

PR #1177, opened 2025-06-11
cli/discover: add 'implicit' config to pair for collection creation

samsonjs:
Adds support for auto-creating collections when they exist only on one side and `implicit = 'create'` is set in the pair config.

Builds off the idea in #869 but ditches deletion since that's not really possible to support in a safe way. Someone else can tackle that if they care.


WhyNotHugo:
Thanks!

Issue #1186, opened 2025-08-27
Show response body when debug logging enabled


raxod502:
Suppose I'm getting an HTTP error from the server DAV endpoints when running vdirsyncer. I then add `-vdebug` as suggested by the error message from vdirsyncer, expecting to see information would tell me the reason for the failure.

Unfortunately, while vdirsyncer logs the request headers and body, as well as the response status code and headers, it does not log the actual response body from the server, which is where the actual error would most likely be found. Instead, it logs this:

```
debug: <StreamReader 339 bytes eof>
```

The log in question is here: https://github.com/pimutils/vdirsyncer/blob/388c16f188a43e3253c6aa56748e035d3159c4e4/vdirsyncer/http.py#L224

I'm not sure how a user is expected to find the error message, except by repeating the HTTP request themselves and seeing what the server returns.

I thought it might be easy to contribute an improvement for this (log the response body on `-vdebug`), but it turns out to be a bit complicated. Firstly, if you just read the body:

```python
logger.debug(await response.content.read())
```

Then it consumes the body and it can no longer be read by vdirsyncer subsequently. It would be necessary to make a copy of the response object, but I could not find any API method for doing this exposed by aiohttp or requests.

Next, I noticed the comment:

https://github.com/pimutils/vdirsyncer/blob/388c16f188a43e3253c6aa56748e035d3159c4e4/vdirsyncer/http.py#L178-L179

I implemented it, following https://docs.aiohttp.org/en/stable/client_advanced.html#aiohttp-client-tracing and https://docs.aiohttp.org/en/stable/tracing_reference.html#aiohttp.TraceConfig.on_response_chunk_received, and added to `vdirsyncer.http`:

```python
async def trace_response_chunk_received(session, trace_config_ctx, params):
    logger.debug(params.chunk)

trace_config = aiohttp.TraceConfig()
trace_config.on_response_chunk_received.append(trace_response_chunk_received)
```

I passed `trace_configs=[http.trace_config]` to the instantiations of `ClientSession` in other modules. However, this did not work. I discovered that `on_response_chunk_received` does not function as documented, as mentioned in https://github.com/aio-libs/aiohttp/issues/5324.

At this point I decided to file this issue with my findings.

WhyNotHugo:
> Then it consumes the body and it can no longer be read by vdirsyncer subsequently. It would be necessary to make a copy of the response object, but I could not find any API method for doing this exposed by aiohttp or requests.

[...]

PR #1187, opened 2025-08-27
Log error response body in debug

raxod502:
See https://github.com/pimutils/vdirsyncer/issues/1186 for a description of the problem. This is a sort of hacky improvement which addresses the usability problem. I don't think it breaks anything, because I realized that seeing the response bodies is most useful when there is an error, and when there is an error, `resp.raise_for_status()` below will throw an exception, so nothing will be able to try reading the already-consumed response body anyway.


WhyNotHugo:
Seems reasonable enough. Thanks.

PR #1211, opened 2026-03-30
Handle missing hrefs in DAV multiget

bernardotorres:
## Summary

Fix DAV multiget robustness when a server omits some requested hrefs in `REPORT` responses.

`DAVStorage.get_multi()` currently raises `NotFoundError` if any requested href is missing from the multiget response. In practice, some DAV servers can intermittently advertise an item and then omit it in a subsequent multiget response. This aborts the whole sync batch.

This patch changes behavior to:

- keep strict `NotFoundError` behavior for single-item lookups (`get(href)` semantics unchanged)
- treat missing hrefs in batch multiget as transient:
  - log a warning
  - continue yielding items that were returned

## Why

In real-world sync runs against Google CalDAV, we observed intermittent omissions for one href in a batch multiget, which caused the whole sync to fail even though other items were valid.

The new behavior preserves strictness where callers expect exact lookup semantics (single href), while improving resilience for batch synchronization.

## Changes

- `vdirsyncer/storage/dav.py`
  - In `get_multi()`:
    - if exactly one href was requested and not returned: still raise `NotFoundError`
    - if multiple hrefs were requested: warn and continue for missing hrefs

- `tests/storage/dav/__init__.py`
  - Added test: batch multiget with one missing href is non-fatal and returns available item(s)
  - Added test: single missing href still raises `NotFoundError`
  - Tests mock DAV `REPORT` using `aioresponses` (matches existing DAV test style)

## Reproduction (before)

`vdirsyncer sync` could fail with:

`Unknown error occurred ... NotFoundError(<missing href>)`

even though the server returned other items in the same multiget response.

## Behavior after

- Batch sync no longer aborts when one href is omitted by the server in multiget
- Single href lookups remain strict and still fail with `NotFoundError`

## Notes

- This is intentionally narrow in scope and only affects missing-href handling in DAV multiget.
- No change to authentication, request construction, or item parsing.

## Validation status

- `pytest` (full suite): **418 passed**, 166 skipped, 2 xfailed.
- `DAV_SERVER=radicale pytest tests/storage -q`: **362 passed**, 71 skipped, 1 xpassed.
- `DAV_SERVER=xandikos pytest tests/storage -q`: **369 passed**, 65 skipped.

### Validation commands and output

```bash
source ~/codes/vdirsyncer/.venv/bin/activate
cd ~/codes/vdirsyncer/vdirsyncer

pytest
# 418 passed, 166 skipped, 2 xfailed

DAV_SERVER=radicale pytest tests/storage -q
[...]

PR #1213, opened 2026-03-30
Refresh SourceHut CI manifests

bernardotorres:
## Summary
- relax the Arch Python guardrail to accept Python 3.13 through 3.14
- move `tests-minimal` to `debian/bookworm` so the minimal dependency set still runs on Python 3.11
- keep `tests-minimal` focused on the minimal dependency matrix by dropping Docker-backed storage tests

## Why
While validating #1211, a few unrelated CI issues surfaced:
- SourceHut no longer provides `alpine/3.19`
- the oldest supported `aiohttp` in the minimal dependency set still needs Python 3.11
- the Docker-backed DAV devkit containers fail with exit status 126 in `tests-minimal`, while storage coverage remains in `archlinux-py313` and `tests-pypi`

### DIFF
--- a/.builds/tests-minimal.yml
+++ b/.builds/tests-minimal.yml
@@ -24,6 +24,7 @@ tasks:
       python3 -m venv $HOME/venv
       echo "export PATH=$HOME/venv/bin:$PATH" >> $HOME/.buildenv
   - docker: |
+      sudo addgroup $(whoami) docker
       sudo systemctl start docker
   - setup: |
       cd vdirsyncer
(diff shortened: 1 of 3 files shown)


.builds/tests-minimal.yml
```diff
@@ -1,30 +1,30 @@
...
   - python3-dev
+  - python3-venv
 sources:
   - https://github.com/pimutils/vdirsyncer
 environment:
   BUILD: test
   CI: true
   CODECOV_TOKEN: b834a3c5-28fa-4808-9bdb-182210069c79
-  DAV_SERVER: radicale xandikos
+  DAV_SERVER: xandikos
```
WhyNotHugo: radicale is not available here?

WhyNotHugo:
Thanks!

PR #1221, opened 2026-06-14
feat(google): add token_command support for external token fetching

fishman:
- allows the use of https://github.com/pdobsan/oama to manage oauth tokens

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -302,16 +302,19 @@ There are two ways to handle OAuth authentication with Google:
    executed to obtain a fresh access token on demand. Vdirsyncer passes the
    token as a ``Bearer`` header and re-runs the command on 401 responses.
 
-With ``token_command``, you do not need to register a Google Cloud project or
-install ``aiohttp-oauthlib``. Any OAuth helper that prints an access token to
-stdout works, e.g. `oama <https://github.com/pdobsan/oama>`_::
+With ``token_command``, vdirsyncer doesn't need OAuth credentials or
+``aiohttp-oauthlib``. The external helper handles Google registration. The format is
+``["strategy", "arg1", ...]``, same as :ref:`the .fetch config values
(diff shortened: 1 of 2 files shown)


docs/config.rst
```diff
@@ -289,10 +278,35 @@ for reference) and the
...
+   executed to obtain a fresh access token on demand. Vdirsyncer passes the
+   token as a ``Bearer`` header and re-runs the command on 401 responses.
+
+With ``token_command``, you do not need to register a Google Cloud project or
+install ``aiohttp-oauthlib``. Any OAuth helper that prints an access token to
+stdout works, e.g. `oama <https://github.com/pdobsan/oama>`_::
+
+    [storage my_google_cal]
+    type = "google_calendar"
+    token_command = ["oama", "access", "my-google-account"]
```
WhyNotHugo: Don't you still need to register a project and provide a `client_id` and `client_secret` to oama?

vdirsyncer/storage/google.py
```diff
@@ -165,6 +205,30 @@ async def _init_token(self):
             # FIXME: Ugly
             await self._save_token(self._token)
 
+    async def _fetch_token_via_command(self):
+        import subprocess
```
WhyNotHugo: Keep this import at the top level.

vdirsyncer/storage/google.py
WhyNotHugo, line 1: We should use the same logic and configuration style as `vdirsyncer/cli/fetchparams.py` for specifying the command; this implementation duplicates a lot of the existing functionality.

WhyNotHugo (changes requested):

docs/config.rst
fishman: Yeah, but not specifically for vdirsyncer. Let me rephrase it

PR #1230, opened 2026-09-11
feat: add Google Calendar scheduling modes

Advil400mg:
Normalize CalDAV scheduling data before upload and add preserve, passive, and standalone behavior for Google Calendar.

Generated with Codex Sol 5.6.


WhyNotHugo:
As a general rule vdirsyncer does not make changes to items: they're synchronise as they are, and it's up to the owner to ensure that the data is in the shape they desire — we don't make assumptions about this.

Advil400mg:
Okay thanks for the feed back.

